PR #83, opened 2014-12-29
pwm.c: cache period

KurtE:
There is an issue that  mraa_pwm_write would fail, because mraa_pwm_read
would fail and return 0.  When the read fails, journalctl shows an
error:
Dec 28 18:01:38 Edison libmraa[365]: pwm: Error in reading period

So now trying version, where the pwm object caches the period, that is
updated whenever you do a read or write of the period.  Side benefit is
that the write should be sped up.

### DIFF
--- a/include/mraa_internal_types.h
+++ b/include/mraa_internal_types.h
@@ -70,6 +70,7 @@ struct _pwm {
     int pin; /**< the pin number, as known to the os. */
     int chipid; /**< the chip id, which the pwm resides */
     int duty_fp; /**< File pointer to duty file */
+    int period;  /**< Cache the period to speed up setting duty */
     mraa_boolean_t owner; /**< Owner of pwm context*/
     /*@}*/
 };
--- a/src/pwm/pwm.c
+++ b/src/pwm/pwm.c
@@ -49,9 +49,13 @@ mraa_pwm_setup_duty_fp(mraa_pwm_context dev)
 static mraa_result_t
 mraa_pwm_write_period(mraa_pwm_context dev, int period)
 {
-    if (advance_func->pwm_period_replace != NULL)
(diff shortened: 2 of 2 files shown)


arfoll:
Thanks @KurtE! Please remember to sign off your commits with `git commit --amend --signoff`.

I have nothing really against this, it seems like a decent idea but the problem is that written period isn't always equal to the read period however that is what calculations will be made against in the kernel. It might be safer to read the period after a write and write that number to dev->period. No?

@tingleby can you review and see my above suggestion? What do you think?

KurtE:
Earlier I did the signoff but it looks like it did not take, so did again, did another push and so hopefully there now.

 

I thought about doing the read back, which could always do or try.  But then was not sure how much it mattered. That is if you ask for a period of 1000 units and it actually set it to 999 and you want a 50% duty, which value do you want it to be calculated against?  Not sure.  Also if it changed your actual period, it could just as easily change your other value, so should you read both back in and make adjustments?  Not sure.

 

My feeling was.  If you want that precise of control of the values you are likely to set to specific value.  Or you can explicitly going to call to Read and then set specific values… 

 

But I am game either way.  

 

I was just glad to figure out why sometimes my TFT display was staying dim, even when I told it to go to 100%.... 

 

Kurt

tingleby:
@KurtE Thanks again for you continued contributions.
The patch looks good to me. Ill give it a quick test and merge it.
 
@arfoll I had to same concerns as you in regard to the period in memory not matching what has been set on the board. Perhaps a mraa_pwm_cache_period would be appropriate when the speed/constance @KurtE was looking for, isnt required.

But realistically this probably is fine for most users so would make sense to be the default behaviour. I think

tingleby:
Closed with 41956b07d65bfaed7e872632a67af385f5fbacc4

PR #87, opened 2015-01-06
gpio: fault on close mini

KurtE:
The gpio_close_pre was put on the wrong board definition.  So calling
close of GPIO pin on mini-board caused a fault

Signed-off-by: Kurt Eckhardt <kurte@rockisland.com>


arfoll:
Merged with 3efb7a03e1e0b0304ce913d386c22461000a4bf0. Thanks for the fix.

PR #90, opened 2015-01-09
x86/MinnowBoard MAX: Adjust gpio pins based on kernel version

warthog9:
This should adjust the pinouts, at runtime, based on the current
kernel that's in use.  The Linux kernel made a change between
3.17 and 3.18 ( specifically ea584595fc85e65796335033dfca25ed655cd0ed )
that changed the numbering for GPIOs.  This obviously breaks everything
for mraa on the MinnowBoard MAX if you have a 3.18 or newer kernel.

Signed-off-by: John 'Warthog9' Hawley <john.hawley@intel.com>

### DIFF
--- a/src/x86/intel_minnow_max.c
+++ b/src/x86/intel_minnow_max.c
@@ -24,6 +24,7 @@
 
 #include <stdlib.h>
 #include <string.h>
+#include <sys/utsname.h>
 
 #include "common.h"
 #include "x86/intel_minnow_max.h"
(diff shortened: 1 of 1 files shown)


arfoll:
Thanks alot for this, I wasn't aware of this change, it seems all gpiolibs implementation start from ARCH_NR_GPIOS (does anyone know how to 'query' this value at runtime?) so we'll likely need to treat this at a more global level at a later date. I've pushed it to the tocompile branch along with a few tweaks to make the code easier to change. I'll merge to master when I get a chance to give this a quick whirl on my minnow, probably monday.

arfoll:
All merged & tested, works a treat on a random 3.19 from git :). Thanks again!

warthog9:
Awesome!  Glad the fix made it in!

PR #95, opened 2015-01-12
arm-devices

michael-ring:
Added new devices:
Raspberry PI A/B/A+/B+
Beaglebone Black B/C
Banana PI (Pro)


michael-ring:
Please have a look at my changes, I am not 100% done yet (need more testing, test hardware arrives tomorrow) but I would like to get some feedback from you guys.

Especially for the beaglebone part, I am not sure if I have not overdone it a little bit by trying to detect the system in so much detail.

tingleby:
Wow, there is a lot here. When I get chance Ill take a look through it all.
One things to point out, please sign off your commits to comply with https://github.com/intel-iot-devkit/mraa/blob/master/docs/contributing.md

arfoll:
@michael-ring this is a lot of work, thank you very much it looks great. I've made a few minor comments, I'm going to have to dust my rpi and give it a go!

The only real issue with your patchset is that it breaks master with every commit. I would prefer a set of commits like this which would mean we could revert/bisect alot more easily: 
1. modify RPI platform def
2. commit for pwm advance hooks for BBB support
3. BBB platform def
4. bananapi platform def
5. example modifications
6. example styling modifications

Does that make sense?

Also commit messages should start with a module/filename for example "rpi: modified rpi support xxx" and be signed off (just add `-s` to `git commit`). If you need any help with this just shout I'm more than happy to help you out with all of this - except the sign off, that you really have to do yourself :)

michael-ring:
Thank you for the review!

I will take your points into account when I prepare the patches for 
submission, I am still searching for a good solution to handle pwm on BBB.

Michael

Am 13.01.15 um 10:06 schrieb Brendan Le Foll:

alexandruradovici:
Hello,

I have a suggestion regarding the Raspberry Pi pin numbering. A lot of people are used to the WiringPi numbering and might want to try to use that with MRAA. I would suggest to have a way to initialise the library with this pin numbering system.

Alexandru

michael-ring:
This is easy to implement, the thing I am not sure about is how to expose this functionality in a clean manner because this is more or less raspi specific. Any thoughts?

tingleby:
This could be exposed in a enum if it something that users really need. The mraa numbering is meant to follow the physical layout whilst the WiringPi layout follows how they are come out of the SoC.

alexandruradovici:
I am not sure exactly what you mean. Could you point out an example?

alexandruradovici:
Another issue that might be useful for the Raspberry Pi is direct memory access for pins. If I'm not mistaking, MRAA uses sysfs for pin access? In your commits I can see only pin mappings, meaning this goes through MRAA sysfs. 

Is there any way to add a switching so that it could go through direct memory access? Maybe even through Wiring Pi? We tested at Wyliodrin this and it is faster.

arfoll:
We support memory access using the advance feature pointers, see Edison for
example. So it can be enabled for RPI too if someone does the work ;)

The wiringpi numbers can be put in an enum to make it easier to migrate
similar to the edison miniboard enums but essentially I see no reason to
not start numbering from 1 or 0. Linux's enumeration of gpios should not be
relied on as shown by the recent kernel changes in 3.18.

Another issue that might be useful for the Raspberry Pi is direct memory
access for pins. If I'm not mistaking, MRAA uses sysfs for pin access? In
your commits I can see only pin mappings, meaning this goes through MRAA
sysfs.

Is there any way to add a switching so that it could go through direct
memory access? Maybe even through Wiring Pi? We tested at Wyliodrin this
and it is faster.

—
Reply to this email directly or view it on GitHub
<https://github.com/intel-iot-devkit/mraa/pull/95#issuecomment-70006133>.

alexandruradovici:
Wiring Pi numbering starts from 0, but has a different pin order https://projects.drogon.net/raspberry-pi/wiringpi/pins/. I have no idea why it is like this.

michael-ring:
I will add an enum for wiringPi today or tomorrowand will push this together with some more changes and a bugfix for Raspberry Pi B+.  
Implementing direct memory access is on my todo list, I first want to integrate some simple devices (temperature sensors, LED displays....) to make sure the basics work fine on the devices I have access to before I dig into making things faster.

deadprogram:
This looks very exciting!

Issue #80, opened 2014-12-23
[API] Add high & low to direction enum in mraa_gpio_dir


arfoll:
@KurtE in https://communities.intel.com/thread/58176?sr=inbox&ru=161660

Something I wonder about maybe handling in MRAA is that in the mraa_gpio_dir call for the MRAA pin associated with GPIO182 (mini board MRAA pin 0),

In that call you open the file: /sys/class/gpio/gpio182/direction

and you either write in or out to this file to set the actual direction, but I have seen examples (actually in the arduino IDE), where instead of writing the value out, you write either high or low to the file, where internally it sets the direction to out and the value to either 1 or 0...

So wondering what would happen if you made a version of mraa that allowed you to pass in either an optional value or a new value to the call like: mraa_gpio_dir(gpio, MRAA_GPIO_OUT_HIGH);

KurtE:
I did a minimalist fix for this, which simply added two new directions you can pass into mraa_gpio_dir. 
the above values: The current stuff is checked in at:
https://github.com/KurtE/mraa/tree/gpio_in_out

This may be sufficient.  Did a simple test program on my mini-breakout board where first I did it the two step method and then did it with the new values.  I had a GPIO pin MRAA 14 go to breadboard, also with 10K PU resistor and LA hooked up.  With the old way saw where the IO pin get pulled low.  With the new way the IO pin does not...

The only questions are: 
a) Do we need to handle it in other cases.  Example if one of board types does a GPIO_DIR_REPLACE I have not updated those platforms to try this.  

b) If a platform does not support this, should more code be added, to  detect if the write of high or low to the direction file fails and then fall back and write out to the direction file and then call off to mraa_gpio_write with 1 or 0... Obviously easy to do...

Thoughts?  If what I did is sufficient, can easily generate pull request.

tingleby:
I plan on pulling this into the mainline in the next few days. Your comments around if a replace hook does not implement the HIGH and LOW aspects of the switch-cast statement don't worry me too much. As long as the default case is present.

KurtE:
Sounds good.  I only tested it on the Edison, but my assumption is the same thing will work on Galileo as in both cases in the Arduino IDE core sources, sysfs.c uses the values "high" or "low" when switching IO pin directions to output.

arfoll:
Closed with #100.

PR #100, opened 2015-01-20
gpio: mraa_gpio_dir allow output to set val

arfoll:
There is an issue that when you take an IO pin that has an external PU
resistor and switch it from INPUT to OUTPUT the state of the pin will go
down before you can set it back high.

Signed-off-by: Kurt Eckhardt <kurte@rockisland.com>


arfoll:
Merged with 20d229168b5fadb2772f7aa259386574b2fc75a8. Closes #80.

PR #101, opened 2015-01-20
Definition of a MRAA FFI for TypeScript language

barais:
The goal is to provide a support for the typescript language. 

To do that, you just need to create a FFI for Typescript. 
This has been done in src/typescript/mraa.d.ts

Next, I just update the set of examples usig the typescript language

### DIFF
--- a/examples/typescript/bmp85.ts
+++ b/examples/typescript/bmp85.ts
@@ -1,0 +1,57 @@
+///<reference path='../../src/typescript/node.d.ts' />
+///<reference path='../../src/typescript/mraa.d.ts' />
+
+/*
+ * Author: Brendan Le Foll <brendan.le.foll@intel.com>
+ * Modfied by Barais Olivier <barais@irisa.fr> for typescript support
+ * Copyright (c) 2014 Intel Corporation.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining
+ * a copy of this software and associated documentation files (the
+ * "Software"), to deal in the Software without restriction, including
+ * without limitation the rights to use, copy, modify, merge, publish,
+ * distribute, sublicense, and/or sell copies of the Software, and to
+ * permit persons to whom the Software is furnished to do so, subject to
+ * the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be
+ * included in all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+ * EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+ * MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+ * NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE
+ * LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
+ * OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION
+ * WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ */
+
+var m = require('mraa'); //require mraa
+
+// helper function to go from hex val to dec
+function char(x) { return parseInt(x, 16); }
+
+var x = new m.I2c(0);
+x.address('0x77');
+
+// initialise device
+if (x.readReg(char('0xd0')) != char('0x55')) {
+  console.log("error");
+}
+
+// we want to read temperature so write 0x2e into control reg
+x.writeReg(char('0xf4'), char('0x2e'))
+
+// read a 16bit reg, obviously it's uncalibrated so mostly a useless value :)
+console.log(x.readWordReg(char('0xf6')))
+
+// and we can do the same thing with the read()/write() calls if we wished
+// thought I'd really not recommend it!
+var buf = new Buffer(2)
+buf[0] = char('0xf4')
+buf[1] = char('0x2e')
+console.log(buf.toString('hex'))
+x.write(buf)
+
+x.writeByte(char('0xf6'))
+var d = x.read(2)
--- a/examples/typescript/AioA0.ts
+++ b/examples/typescript/AioA0.ts
@@ -1,0 +1,35 @@
+///<reference path='../../src/typescript/node.d.ts' />
+///<reference path='../../src/typescript/mraa.d.ts' />
+
+/*
+ * Author: Dan Yocom <dan.yocom@intel.com>
+ * Modfied by Barais Olivier <barais@irisa.fr> for typescript support
+ * Copyright (c) 2014 Intel Corporation.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining
+ * a copy of this software and associated documentation files (the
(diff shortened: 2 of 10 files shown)


examples/typescript/bmp85.ts
```diff
@@ -0,0 +1,57 @@
...
+ * NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE
+ * LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
+ * OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION
+ * WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ */
+
+var m = require('mraa'); //require mraa
+
+// helper function to go from hex val to dec
+function char(x) { return parseInt(x, 16); }
```
rwaldron: This is unnecessary considering the use case is to pass a string that looks like a HexIntegerLiteral, where a real HexIntegerLiteral would do just fine.

```js
0xd0 === parseInt("0xd0", 16);
```

examples/typescript/bmp85.ts
```diff
@@ -0,0 +1,57 @@
...
+
+var x = new m.I2c(0);
+x.address('0x77');
```
rwaldron: Does this actually work? I've never tried a string here, but the api expects a uint8

examples/typescript/bmp85.ts
```diff
@@ -0,0 +1,57 @@
...
+
+// initialise device
+if (x.readReg(char('0xd0')) != char('0x55')) {
```
rwaldron:
```js
if (x.readReg(0xd0) !== 0x55)
```

examples/typescript/bmp85.ts
```diff
@@ -0,0 +1,57 @@
...
+  console.log("error");
+}
+
+// we want to read temperature so write 0x2e into control reg
+x.writeReg(char('0xf4'), char('0x2e'))
```
rwaldron:
```js
x.writeReg(0xf4, 0x2e);
```

arfoll:
Hi @barais, thanks for the work & your interest in mraa! I'm honestly a little confused as to what you are doing and the reasons behind it, could you explain a little more? I've got a small list of queries - excuse my lack of knowledge until you made this PR I was unaware of 'typescript' :)

1. The .ts files seem to reuse the nodejs examples, why are both necessary?
2. Is the mraa.d.ts autogenerated? Does it generate from the mraa js module?
3. Since this requires the nodejs module which is automatically generated - can we automatically generate the dts?
4. What is the node dts, that looks like something which is always required, why should we distribute it?

@rwaldron These examples all work on my box on nodejs - feel free to open a new PR or issues if you think we can do it in a better way, the parseInt() call may have been superfluous I was being fast and got annoyed at node's constant misunderstanding of my hex. This is probably not the place to discuss these issues though - unless specificaly relevant to typescript.

@barais If it's easier feel free to pop into #mraa on freenode IRC and chat with us (invitation is open to everyone)

examples/typescript/bmp85.ts
barais: Thanks for reviewing this code. You right, I go too quickly on when I rewrite the example. 

I will correct that.

rwaldron:
>  These examples all work on my box on nodejs

How does `x.address(...)` know to change the string argument that it's receiving to a number?

> the parseInt() call may have been superfluous I was being fast and got annoyed at node's constant misunderstanding of my hex.

As long as you're writing HexIntegerLiterals the correct way, eg. `0x00`, then there is no reason that "node" will misunderstand. Additionally, "node" has nothing to do with the actual execution of code—this is done by v8.

> This is probably not the place to discuss these issues though - unless specificaly relevant to typescript.

I saw them here for the first time—I didn't realize that the JS files had all been changed to be written that way.

barais:
@rwaldron 
1. The .ts files seem to reuse the nodejs examples, why are both necessary?  
> I can create example that use class to be more on the typescript spirit. 
2. Is the mraa.d.ts autogenerated? Does it generate from the mraa js module?  
> Partialy but I can improve my generator to fully generate the d.ts if you prefer for supporting evolution. 
3. Since this requires the nodejs module which is automatically generated - can we automatically generate the dts?  
> I don't think so
4. What is the node dts, that looks like something which is always required, why should we distribute it?  
To be sure that the examples works directly. But we can also provides something in a Readme.  

The idea behind this work is that f some developper, especially the one that comes from Java or C#, it is a way to get something familiar. It also provides also refactoring capabilities.

arfoll:
@barais Thanks for explaining a little, it's very helpful.
1. The examples are meant to show a minimal working set for working with the library, if the js examples already work I see no reason to provide more examples. The less code is around, the easier it is to maintain :)

2. Honestly fully generating the d.ts is likely the only way this would get into master, whilst we endevaour to make little change to libmraa maintaining both a C++, C and SWIG APIs all in sync is hard and having a d.ts file to update would mean we'd likely end up with code rotting.

3. I dont think your i2c or spi works because from what I see in the d.ts you don't do the correct translations as we do with SWIG.

To be honest I'm struggling to understand whether this should be included, to get C# or java support we can use SWIG to generate bindings which would be easier to support and fits in with our existing model, we've done it with PHP & go with little trouble and at least java is very well supported in SWIG. What's the advantage of using typemaps to do it instead?

@rwaldron ah yes the address() param as modified there doesn't work. I had some issues with HexIntegerLiterals in node, I'll have to try again.

barais:
> Honestly fully generating the d.ts is likely the only way this would get into master, whilst we endevaour to make little change to libmraa maintaining both a C++, C and SWIG APIs all in sync is hard and having a d.ts file to update would mean we'd likely end up with code routing.  

I fully agree, I will provide this generator from .hpp header file or swig interface files. 

> I dont think your i2c or spi works because from what I see in the d.ts you don't do the correct translations as we do with SWIG.

I will correct that. 

> To be honest I'm struggling to understand whether this should be included, to get C# or java support we can use SWIG to generate bindings which would be easier to support and fits in with our existing model, we've done it with PHP & go with little trouble and at least java is very well supported in SWIG. 

At the end, it is not for developing with C# or  Java but with TypeScript. Typescript is also a language which is used by javascript and nodejs community and it requires this d.ts file. 
 
> What's the advantage of using typemaps to do it instead?

basically, it is just typing information for typescript compile. At the end, it directly use the js library you already provide. 

I propose you  reject this pull request. I will rework on it and provide the d.ts generator. Next time I will contact you on IRC before, to see if it is valuable for the community. If it is not I will keep it on a public repository. 

Thanks for your work and your quick review

Regards

arfoll:
Ok, I think I understand much better now, it provides static typing and object handling as a superset to js by means of a d.ts wrapper/header, makes sense. Do you have a link to your generator? I'd be very interested in playing with it if it's publicly available :) Because our code is actually C++ we have the typing information already, by using the output from SWIG you should be able to gather all the information you need to generate such a header on the fly, which would be quite cool.

Sure, I'll close this for now, but I'd be really interested to hear from you again so do drop by #mraa. I think something like this would be valuable to the community, we have a very wide range of developers from a lot of backgrounds so providing more choice is always a good thing!

rwaldron:
> it provides static typing and object handling as a superset to js by means of a d.ts wrapper/header, makes sense.

But only at "development time". All of that is stripped away before the code is actually run, which leaves it at "just JavaScript".

PR #106, opened 2015-01-25
mraa: we now log version and user ID into syslog upon init

alext-mkrs:
How about such a solution to #98.

If we anyway want to log something (version), then let's reuse the existing DEBUG-level message, convert it to NOTICE and add version there.

It looks like that in the log:
```
Jan 25 14:36:33 edison-sd libmraa[493]: libmraa version v0.5.4-100-g6c73a8a initialised by user 'root' with EUID 0
```


tingleby:
Closed with 82bb1c5

PR #111, opened 2015-02-01
aio.c: added normalized read returning a float in a 0.0-1.0 range

alext-mkrs:
Here is my take at #32 as far as I understood it.

@arfoll, @tingleby, I'd be grateful for comments on the code itself as well as if I got the request correctly at all :smiley:

Feel free to assign this one to me, I'll drive it to closure even if this first attempt is not really what we want to have here.

Tested with C, JS and Python AIO examples, works as intended, e.g. in JS it looked like:
```
<cut>
MRAA Version: v0.5.4-105-gbaa1a0a
analogvalue is: 822
analogValueFloat is: 0.80352
MRAA Version: v0.5.4-105-gbaa1a0a
analogvalue is: 996
analogValueFloat is: 0.97067
MRAA Version: v0.5.4-105-gbaa1a0a
analogvalue is: 1023
analogValueFloat is: 1.00000
<cut>
```

### DIFF
--- a/api/mraa/aio.h
+++ b/api/mraa/aio.h
@@ -68,6 +68,14 @@ mraa_aio_context mraa_aio_init(unsigned int pin);
 unsigned int mraa_aio_read(mraa_aio_context dev);
 
 /**
+ * Read the input voltage and return it as a normalized float (0.0f-1.0f).
+ *
+ * @param dev The AIO context
+ * @returns The current input voltage as a normalized float (0.0f-1.0f)
+ */
+float mraa_aio_read_normalized(mraa_aio_context dev);
+
+/**
  * Close the analog input context, this will free the memory for the context
  *
  * @param dev The AIO context
--- a/api/mraa/aio.hpp
+++ b/api/mraa/aio.hpp
@@ -66,6 +66,14 @@ class Aio {
             return mraa_aio_read(m_aio);
         }
         /**
+         * Read a value from the AIO pin and return it as a normalized float.
+         *
+         * @returns The current input voltage as a normalized float (0.0f-1.0f)
+         */
+        float readNormalized() {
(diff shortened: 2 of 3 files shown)


tingleby:
Great,
I've added comments to the patch, please take a look.

alext-mkrs:
Fixed the function name in C++ header.

By the way, it would probably be worth it to update the examples as well, what do you think?

If yes, should that be a separate commit (as in "actual library code change vs change in examples") or should I squash that into a single one (as in "this is logically a single change, which adds new functionality and examples of using it")?

arfoll:
@alext-mkrs Thanks alot, sorry for the slow response I've been unwell the last few days. How annoyed would you be if I renamed normalized to float? It's shorter and there is only one 'correct' way of spelling it. If you don't mind then I'll merge this in (don't worry about the rename i'll handle that).

Personally I think 2 patches is best since one is adding an API one is simply giving an example of how to use it and it means its easier to trace when things go wrong when things are more split.

alext-mkrs:
@arfoll I'd be absolutely fine with that :smiley: I actually named it "_float" at first, but then decided it would make sense to get away from type names (as they could change over time and can be different in different languages) and changed it to "_normalized" before sending the PR.

I can rework that - no problem, I'll also add examples in a second commit.

Get well soon!

arfoll:
Don't worry done for you. cd6701d604e531cbfa6fbf163c507d94d6566df7. I also removed a temp var you had in the read_float() call.

I was about to push it, then started typing 'normalized' realised that was hard to spell and thought something else would be better :) _float is not necessarily the best but I can't think of anything with is short-ish.

alext-mkrs:
Ok, thanks :smiley:

PR #118, opened 2015-02-10
Add options for nodejs include directory

yoneken:
As there is a few options of nodejs include directory, we should modify the CMakeLists.txt.

In ubilinux (Debian), node.h is located at /usr/include/nodejs/src/node.h by aptitude.

Signed-off-by: Kenta Yonekura <yoneken@ieee.org>


arfoll:
Thanks alot for this, I still think the header on ubilinux is in the wrong place but I guess I have to live with it :) Merged with 2965665feee80e4f31de8da977a771159bb4e4d1

PR #120, opened 2015-02-12
Install mraa python lib gently

yoneken:
This modification provides a gentle way to install the mraa python library.

* Installer searches a proper site-packages directory
* Users can uninstall the library with pip.

Signed-off-by: Kenta Yonekura <yoneken@ieee.org>


arfoll:
This looks really great thanks, I'm waiting to try this on a few more boxes make sure this doesn't break builds with venv or python2/python3 boxes etc...

arfoll:
This breaks the cmake module for generating ipks, I've pushed it to the tocompile branch. @tingleby can you take a look? It would be nice to have this in.

This change also requires python-setuptools to be added to images, I've done that on devkit1 schroot.

arfoll:
This is really nice but for now I'm going with #141 as a solution which also works with cross compiling our opkg infra and python2/python3. I would however be interested in submitting a pkg to pip in a similar way to npm so if you have an idea I'm open to suggestions.

Thanks!

PR #123, opened 2015-02-14
examples/gpio: Extend gpio sample

michael-ring:
I have extended the gpio example so that it is able to also do mmap access to gpio ports. 
I have also added some whitespace to list command so that there is vertical alignment


arfoll:
Thanks Michael, merged.

PR #124, opened 2015-02-14
SPI fixes for Beaglebone Black, New Example for SPI

michael-ring:
I have added Initialization for SPI, reason is that on Beaglebone Black SPI fails when not properly initialized.
The example shows how to use 16 bit transfers, this is usefull for devices like the max7219 that work with SPI in 16 bit mode. The 16bit mode should also make it possible to use 9Bit mode on Raspberry Pi, but I did not test this.

### DIFF
--- a/api/mraa/spi.h
+++ b/api/mraa/spi.h
@@ -104,6 +104,14 @@ mraa_result_t mraa_spi_frequency(mraa_spi_context dev, int hz);
  */
 uint8_t mraa_spi_write(mraa_spi_context dev, uint8_t data);
 
+/** Write Two Bytes to the SPI device.
+*
+* @param dev The Spi context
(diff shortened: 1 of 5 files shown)


arfoll:
Can you check your commit prefixes? They're all examples/spi but that's not right, I can fix as I merge if it's easier.

I'll need to find a way to use these _word funcs from js/python :)

michael-ring:
Can you please fix this, I knew that I did that wrong but could not find an easy way to fix that.

arfoll:
Sure, not a problem.

arfoll:
Rebased and pushed with ffe40033d4b03d8ceb227aa092142fc08be16ae8 (and it's friends). Thanks alot!

Issue #113, opened 2015-02-01
Extended SPI Functions 


michael-ring:
I have implemented an example for the MAX 7219 Chip.

This chip uses 16Bit SPI which causes some problems on Raspberry Pi and possibly other devices because their SPI driver only supports 8/9 Bits. 

To make this chip work I would like to implement an extended mraa_spi_bit_per_word function that takes a GPIO Pin as an additional parameter. This GPIO Pin then takes over the functionality of CS and makes it possible to have 16/32 Bit transfers on those limited devices. 

I have also already implemented helper functions to send 16/32 Bit values via SPI, those functions and the mraa_spi_transfer_buf can get modified to support this extra CS seamlessly.

Any objections?

michael-ring:
Implemented in Pull Request #125

Issue #112, opened 2015-02-01
SPI Interface not fully initialized on mraa_spi_init


michael-ring:
I found out that on the Beaglebone I need to first set the spi Interface Mode before the interface works correctly.

There are now two possible ways to solve this, either

initialize spi as master per default on mraa_spi_init 

or

change the examples so that they always include a call to mraa_spi_mode

I can provide patches for both, I have no access to galileo/minnow/nuc board so I cannot test if it hurts to do set a default spi mode on init.

I can provide patches for both versions.

tingleby:
Thanks Michael!
Out of the two I think "initialize spi as master per default on mraa_spi_init" would be a better fit.
Please go ahead and provide a patch, I will test it out on at least the galileo/edison.

michael-ring:
Will do.

Could you please reject my previous pull request, I would like to create a new pull request and take into account the tips I received for the old one. I found out that with the github UI it is only possible to handle one pull request to one repo, there seem to be workarrounds for this but I think it is easier to trash the hole request and to start over with more atomic patches.

tingleby:
Are you talking about #95 ?
If you were to submit the PR from a branch then you can have as many PRs as you feel like. If you take a look at AlexT's #111 it has been submitted from a "feature" branch.
Just found this link to try and provide you a rough guide, http://blog.milkfarmsoft.com/2011/03/on-feature-branches/

michael-ring:
Thank you, I will have a look at this

michael-ring:
Fixed in pull request #125

arfoll:
Closed as this is no longer an issue as I understand?

PR #125, opened 2015-02-14
examples/i2c_ht16k33.c Example for the ht16k33 LED driver chip

michael-ring:
Signed-off-by: Michael Ring <mail@michael-ring.org>

### DIFF
--- a/examples/CMakeLists.txt
+++ b/examples/CMakeLists.txt
@@ -10,6 +10,7 @@ add_executable (mmap-io2 mmap-io2.c)
 add_executable (blink_onboard blink_onboard.c)
 add_executable (uart_setup uart_setup.c)
 add_executable (gpio gpio.c)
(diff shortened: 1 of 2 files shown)


michael-ring:
This is a small sample app that switches LED's on a ht16k33. This chip is used on quite some adafruit breakout boards

arfoll:
I'm a bit confused about why you referenced #112 & #113, this is a SPI sample? Is there any reason to have this in here instead of a module in UPM?

michael-ring:
Sorry, this reference was unintentional, the i2c code has no dependency whatsover to the spi code.

This module is test code so that I can test i2c with real hardware on different platforms, testing is easier when the module is within mraa, but I can live with it when you do not want to merge this file into mraa.

arfoll:
As discussed, we already have an i2c sample so this would be duplication. In fact we have 3 python i2c samples... I'm not very good at following my own rules apparently!

PR #126, opened 2015-02-15
Extended support for raspberry Pi

michael-ring:
The support for Raspberry Pi is now complete for all popular models, only _real_ support for the compute model is missing as I do not own one of these and do not intend to buy one now that the Raspberry Pi 2 is available.


arfoll:
This looks great to me, can you fix the last patch with the comments (if they make sense) and I'll merge it.

michael-ring:
Done..

arfoll:
Merged, this is really great! I've also updated the docs to reflect the latest changes, though they could do with a bit more love. Now to dust my rpi and try this out for real :)

PR #133, opened 2015-02-18
mraa_uart_get_dev_path: return NULL on error

Jon-ICS:


arfoll:
Merged with dd807308ac7a8a87454a4c8cdf9dbe287ce9275b. Thanks!

PR #134, opened 2015-02-18
mraa_gpio_wait_interrupt: check fd BEFORE attempting to use it

Jon-ICS:
In addition, an fd equal to 0 is perfectly legal.

### DIFF
--- a/src/gpio/gpio.c
+++ b/src/gpio/gpio.c
@@ -151,6 +151,10 @@ mraa_gpio_wait_interrupt(int fd)
     unsigned char c;
     struct pollfd pfd;
 
+    if (fd < 0) {
+        return MRAA_ERROR_INVALID_RESOURCE;
+    }
+
     // setup poll on POLLPRI
     pfd.fd = fd;
     pfd.events = POLLPRI;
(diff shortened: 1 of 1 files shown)


arfoll:
You're entirely right 0 is valid but isn't it always stdin?

Nex time please split this type of thing into 2 patches, one where you move the function and another where you change the if statement. I'll merge this once I've given it a quick test on a board.

Jon-ICS:
On 02/19/2015 03:56 AM, Brendan Le Foll wrote:
>
> You're entirely right 0 is valid but isn't it always stdin?
>


You can close stdin, stdout, and stderr if you have no use for them - 
I've actually done this for software which runs as a daemon at system 
startup for example.


> Nex time please split this type of thing into 2 patches, one where you 
> move the function and another where you change the if statement. I'll 
> merge this once I've given it a quick test on a board.
>

I did not move the function, only the if statement up a few lines... Or 
am i missing something?

-- 
Jon Trulson
Integrated Computer Solutions
www.ics.com

arfoll:
Commits should only ever do 1 thing :) Yours does 2. That's all. I split it into ee05b49ea2e23ebe4e5c0cde96bd5dba6d593570 & 9d80b1252152915c617dd22ca09f61e28f844279. Thanks!

PR #140, opened 2015-02-27
I2C functions to read multiple bytes from a register offset

Yoda12999:
These functions allow for a write command to be sent with the address to start from and then a read command without a stop command in between the two.
There is also a commit that fixes some very small formatting errors.

### DIFF
--- a/api/mraa/i2c.h
+++ b/api/mraa/i2c.h
@@ -116,6 +116,17 @@ uint8_t mraa_i2c_read_byte_data(mraa_i2c_context dev, const uint8_t command);
 uint16_t mraa_i2c_read_word_data(mraa_i2c_context dev, const uint8_t command);
 
 /**
+ * Bulk read from i2c context, starting from designated register
+ *
+ * @param dev The i2c context
+ * @param command The register
+ * @param data pointer to the byte array to read data in to
+ * @param length max number of bytes to read
+ * @return The length in bytes passed to the function or 0
+ */
+int mraa_i2c_read_bytes_data(mraa_i2c_context dev, uint8_t command, uint8_t* data, int length);
+
+/**
  * Write length bytes to the bus, the first byte in the array is the
  * command/register to write
  *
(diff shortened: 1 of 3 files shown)


arfoll:
Thanks for the interest @Yoda12999 ! It's funny because I was talking about something like this only yesterday :)

Few general comments:
1. You need to sign off your commits as explained in docs/contributing.md
2. Try remove trailing whitespace, a quick `git log -p` should highlight them in red
3. The styling fix shouldn't move the pointer indicator, since that's (for better or worst) how 'most' of the code is written :). If you have a good argument for moving them then I'm easy I just think it's nice to keep it clear what the 'type'.

arfoll:
To give a rough idea of the changes here's the diff I had on your patch - http://dpaste.com/3XG7ZW3

Yoda12999:
I'm sorry that I didn't sign off (or check my code in the morning before committing, late night coding is never a good idea), I will make sure to do that in the future. Regarding the pointer indicator, when looking at the few in i2c.hpp they were placed on the the name rather than the type. I see after looking into it further that this is the exception rather than the norm.

Regarding the return type, I would have rather had it return the length of the read done, but I am pretty sure that a ioctl RDWR call does not return this information. I am pretty sure from looking at other code that it returns 0 on success and -1 on failure. It could be changed to return 0 on failure and the length passed on success, and that may be the closest to other read functions that can be achieved.

arfoll:
@Yoda12999 no need to apoligise - thx for using your nights to send us patches :) Do you want me to rebase all those togehter or are you ok to do it youself and push force to your repo?

The reason I want it to return the length is that it's this mechanism that enables the python/nodejs modules to work :) Functions in the C API typically return a mraa_result_t (0 being MRAA_SUCCESS).

About the code style, it's surprisingly hard to not let such minor things slip in the codebase, we really could do with an autoformating check :)

Yoda12999:
There you go, that should be everything in one commit.

arfoll:
Merged with caf13833252ff0fa869d225a118779b3504ab51e. Thank you very much! Also merged the tab/whitesapce fix with 0dd03950e41e0b9024b08238ed90e1d1c535099d

PR #141, opened 2015-03-01
build with python 2.7 in the correct python site dir

tripzero:
Right now the python bindings are not being installed in the correct site dir.  It tries to make a guess which may not be correct depending on the distro.  This patch fixes that by getting the site dir from the python interpreter.

This also hardcodes mraa to build against python 2.7.  It's possible to override this, but currently python3 interpreter detection is broken in cmake < 3.1.

### DIFF
--- a/src/python/CMakeLists.txt
+++ b/src/python/CMakeLists.txt
@@ -1,4 +1,12 @@
-FIND_PACKAGE(PythonLibs)
+SET(PYTHONBUILD_VERSION 2.7)
+
+IF(BUILDPYTHON3)
+  SET(PYTHONBUILD_VERSION 3)
+endif()
+
+FIND_PACKAGE(PythonLibs ${PYTHONBUILD_VERSION} REQUIRED)
+FIND_PACKAGE(PythonInterp ${PYTHONBUILD_VERSION} REQUIRED)
+
 INCLUDE_DIRECTORIES(${PYTHON_INCLUDE_PATH})
 
 include_directories(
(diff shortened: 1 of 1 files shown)


tingleby:
Im not entirley sure about running the python interpreter to get the site dir but if it works everywhere we already build mraa, there shouldnt be a reason why not.
In regards to the patch in itself, please add a sign-off. See https://github.com/intel-iot-devkit/mraa/blob/master/docs/contributing.md

tripzero:
It will add a dependency on python... but it should work.

arfoll:
This is quite cool but I'm not sure I like this method over PR #120. True that #120 means a dep on python setuptools. The problem with either of these methods is when cross compiling and using the cmake IPK package build scenario. I'm tempted to say #120 should be default and when BUILDIPK is set we just do the old behaviour. Let me have a think and get back to you on this.

tripzero:
I didn't know PR #120 existed.  Having a setup.py seems like a typical way to install python packages, however, I'm not sure how standard that is in mixed projects with build tools like cmake...  My PR comes from findings on what other cmake projects have done to install python bindings.

tripzero:
Any additional thoughts on this?  I have other patches in the work I don't want to mix with this if this will be abandoned.

arfoll:
Well honestly your method breaks cross compiling and the other method breaks BUILDIPK, so I'm still working on a solution for everyone :) It's likely I'll take something like your method with a bunch of other bits to do both python2 & 3 builds that I've been working on but haven't quite had time to polish with other stuff coming up - hopefully this week I get a chance to push all that stuff out.

If you have other patches in the pipeline I recommend you send them from different branches as long as they don't need to go on top....

arfoll:
Merged - with a few modifications so that it works in Arch Linux & Poky (YP) as well as debian

Issue #149, opened 2015-03-03
[edison] spi kernel driver does not let you disable CS


arfoll:
Using the NOCS flag does not work on edison spi kernel driver.

```
root@edison:~# ./spidev_test -D /dev/spidev5.1 -s 1000000 -N
can't set default spi mode: Invalid argument
Aborted
```

We should be able to implement something like this - http://dpaste.com/1N276PZ to disable HW CS completely.

arfoll:
Closing as hardware issue

Issue #132, opened 2015-02-17
SPI module, setting frequency imposes no limit


cosscat:
Hello,

While testing SPI of mraa on Gallileo v1, I observed that there is no limit for setting the frequency in Hz except the boundries of integer values. Is there a way to interogate the hardware for the max. frequency it accepts?

>>> m.getVersion()
'v0.5.4-118-g818e1a8'

>>> s.frequency(1000)
0
>>> s.frequency(1000000)
0
>>> s.frequency(10000000000)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python2.7/site-packages/mraa.py", line 795, in frequency
    return _mraa.Spi_frequency(self, *args)
TypeError: in method 'Spi_frequency', argument 2 of type 'int'
>>> 

Regards,

Costin

arfoll:
Yes, we probably should have a MAX limit for spi bus freq, we currently don't and just accept anything that spidev will let us set.

michael-ring:
I think there are two problems here.

First is that python is called with something that is most likely too big for an int. (No idea if this is really the case, me not python)

the second problem is that the frequency value can be too high for the spi interface, this problem should be fixed with latest commit, you should get a warning in syslog and the maximum is set to what the spi says is max.

raoloka:
On the other hand, even though the function is called to increase the frequency, it doesn't seem to be
effective. I noticed the clock frequency still hovered around 4MHz.

alext-mkrs:
Looks like we can close this one - the original issue is fixed by @michael-ring in his commit 3a5b99252419b5, so we now check the driver for the max SPI frequency.

We also have a hardcoded value in `mraa_spi_init_raw()`, which caps the frequency to 4MHz, even if driver tells us it can support more. And that magic value was there since the very beginning of the SPI support in mraa, as far as I can see. @tingleby, is there any specific idea behind the value (you were the author of that first commit :smiley:)?

What's interesting is that we don't do the same hardcoding in `mraa_spi_set_frequency()`, we only interrogate the driver - which looks like an inconsistency to me. Should we remove the hardcoded value from `_init_raw()`, maybe?

alext-mkrs:
@tingleby, ping - any comments on the above?

tingleby:
The cap of 4Mhz was probably a limitation of the original board, Galileo Gen1. If you can get the clock higher than that, then by all means.

alext-mkrs:
[...]

Issue #61, opened 2014-12-09
[API] Support writing on SPI bus in anything other than 8bit


arfoll:

michael-ring:
Please check #124, this should extend SPI support to 9 and 16 bits, those two are the methods of transfer that usually get used afaik.

Or were you thinking of manually triggering CS, this could come handy to support 16 bit transfer on raspberry pi and other devices with rather limited spi drivers

PR #153, opened 2015-03-08
Soft-SPI Implementation

### LINKED ISSUE #147: SPI transfer with 16 bits per word for platforms that do not support this in hardware
I have implemented and tested code to enable support for 16bit transfers with the Raspberry Pi. There is still support for some cases missing (lsb first and MODE1) but so far this works really nice.
I would like to add checks to spi_write and spi_write_word that they refuse to send data when the numer of bits set does not match the size of the data. example:

mraa_spi_write should refuse to work when bits_per_word is >8
and
mraa_spi_write_word should refuse to work when bits per word is <9 and >16

Not doing so will create quite some special cases that need to get handled and code would be cleaner as cases like a user wanting to send 3 bytes in 16bit mode simply would not exist.

Are you guys OK with those changes?

Michael

michael-ring:
Soft SPI implementation helps platforms that have drivers for SPI that only support 8/9 bit transfers to do 16bit transfer by replacing the driver's Chip Select with a chip select created by mraa lib.

Parts of the code are based on an example provided by arfoll in #149 

This is the pull request to fix #147, I think the other SPI related issues: 
#61 #113 #132 can also get closed because they were covered by older pull requests.


michael-ring:
btw, I messed up my commits a little, hope this does not cause a problem for the merge, the files changed list looks good to me.

arfoll:
Thanks @michael-ring, I'll have a look at this seriously tommrow but dw about the merge, I'll clean that up on my end.

KurtE:
As I have mentioned in some previous threads, I have been using software CS pins with SPI to handle some graphic displays, like the Adafruit TFT display (ILI9341), which has a CS pin for the display as well as a CS pin for the touch panel.  In addition it has another pin DC(Data/Command) that needs to have it's value synchronized with the SPI pins.  

So far I have been able to deal with it by, simply creating a MRAA gpio pin for the hardware CS pin after I create the SPI object.  This switches the underlying pin back to Mode0, where I am in control of the actual output. 

I was also experimenting with my own branch that optionally did not setup hardware CS pin in MRAA.

In particular for the Edison, you had a flag, like what was added for this pull, but then in:
mraa_intel_edison_spi_init_pre(int bus) 
If that flag was set you would not make changes to the hardware CS pin.

Note: recently a change was added that looks like CS is not touched for mini board, but it looks like it is for the Arduino board (pin 10...)

arfoll:
I've changed a few things and merged everything into one commit - can you check here this does what you want?

https://github.com/intel-iot-devkit/mraa/commits/tocompile

oloynet:
Great news ;)

As I can understand, the CS will be still down only for the "softbits"
value (8, 9, or 16 bits).

Olivier




2015-03-17 16:14 GMT+01:00 Brendan Le Foll <notifications@github.com>:

michael-ring:
Correct. You have to consider SCLK and the number of clock cycles cannot be changed (afaik). So you can only support multiples of the bit values that the platform supports because you cannot reliably put CS high after (for example) 5 SCLK Cycles. So it might happen that 6 or only 4 bits are shifted into your device, not good at all ;-)

As 16 bit is quite often used this is also completely transparent to use, use mraa_spi_sw_cs in this case, all the cs switching will be handled by mraa, you can select any available GPIO as CS pin.

For 24,32,... bits you simply turn off hwcs completely with mraa_spi_hw_cs and send 4x8 bits of data and handle cs switching by yourself with any GPIO port you like.

KurtE:
I am probably missing something, but I am having a hard time trying to figure out when I would use the Software CS.  Don’t get me wrong, I need software control over the CS pin and with most of the SPI based TFT displays, have to do the CS manually.   So in my case I wish to turn off Hardware CS.

 

But for example: in the call mraa_spi_init,  it will call advance_func->spi_init_pre(bus), which in this case is: 

mraa_intel_edison_spi_init_pre: And in the case of the Arduino breakout board this is going to: Change pin 10

(OUTPUT, PULLUP…)

 

And then it will return to mraa_spi_init, where it will see that the CS for this SPI bus has mux information and it will switch pin 10 to mode 1, into hardware SPI mode. 

 

So if you are adding support to logically turn off hardware CS support, you should update these functions to not change the IO pin that is associated with the hardware CS.

 

Personally this is something I would like to avoid.  That is take my Adafruit 2.8” tft display with resistive touch screen.  There are two SPI objects here. 

The first is the display(ILI9341), which uses normal SPI pins including pin 10 for CS, but in addition uses pin 9 for DC (data/command).  

 

In addition it has a second SPI device (STMPE610) which uses pin 8 as the chip select.   

 

In theory I could potentially try to use the Hardware CS for the TFT, however currently not enough control over it.  The ILI9341 display sometimes uses the CS to know when a command has completed.  For example reading in pixel data. If the CS is lost, it will no longer return data.   SPIDEV does have a control fieldspi_ioc_transfer. cs_change  that I believe is supposed to do this, but don’t see any way in MRAA to use it, nor have it I tried it on the Edison to see how (if at all) it works within spidev. 

 

Again sorry if I am missing something here.

 

Kurt

michael-ring:
For the Edison the use of softspi is most likely very limited because the SPI driver is very flexible. On other devices the limitations are more severe, Raspberrry Pi and Banana Pi have a very limited driver (see comments above) and on the Beaglebone Black there exiasts a nice expansion board from mikroe.com that allows you to interface to a lot of SPI/I2C boards the company produces. 

Unfortunately they have not wired the CS pins to the 'official' CS pins for the Beaglebone but instead they use some free GPIO-Pins for CS. 

In those cases the softspi is very helpful, it allows you to work on those devices without the need to change your main code much, all the details are handled in mraa, you only have a different initialisation.

I hope this explanation helps.

KurtE:
Thanks, 

 

That helps. 


Actually as for SPI being flexible on Edison?  Currently defined 2 hardware CS pins, 1st taken by ADC even on those boards (like mini, like sparkfun, like Trossen), who are not the Arduino board, where it is supported… The SPIDEV object is appears to be actually a standard IO pin that is handled at the kernel level to emulate CS pin.  Not sure if more could be defined?  

 

Performance: Right now just about every other platform completely outperforms the SPI of the Edison, many by orders of magnitude.  For example the Raspberry Pi supports the ILI9341 and touch screen at kernel level (pitft) and I believe can get 20-30 frames per second video.  Likewise for other Linux boxes like Odroids (FBTFT).  With Edison I think we are getting maybe .75 frames per second.  (Sorry maybe a little off subject)

 

But again if you are adding a logical CS pin to be handled by MRAA instead of “hardware” CS pin, including turning off the hardware CS (Is that IOCTL supported by Edison?).  I believe that when you say no hardware CS pin, that the actual  default (or only) hardware CS pin should not be updated.  

 

Also I believe that MRAA should be updated to have at least one of the transfer functions that allow me to pass in should the CS pin be modified or not.  However I have seen two different ways that spi_ioc_transfer. cs_change  is used when I look online.  Some appear to imply that when selected, this will deselect SPI  before  the transfer.  It also sounds like the field delay_usecs is used to determine how long between bytes before optionally deselecting.   I read that as if you do not output something again within delay_usecs the system should deselect CS…. Not sure if anyone implements that or not.  

Personally I just wish I could say hold the cs after this transfer… But again this may be slightly out of context to your update. 

 

Thanks again

sriranjanr:
Hello,

I also want to use a custom chipselect.I want the CS to be high initially just before the transfer and go low with the first falling edge of the SCK first bit transfered. How do I achieve this?
My CS (yellow) and SCK (green) should be as in the following diagram.
img_20150801_151815

arfoll:
Closing as very outdated, and soft CS is not something we can seriously consider for a good majority of platforms. The new edison SPI driver is much better than the old one, so there is at least some improvement there.

PR #163, opened 2015-03-11
Fix cpack package architectures and added option for installing gpio tool

whbruce:
Currently package architecture names are incorrect for debian (is x64_64, should be amd64) and for Yocto ADT (is i386, should be derived from traget architexture and tuning e.g. core2-32).

Also added option to install gpio tool as I use it all the time.

Sorry - should have ben working on a branch. Got lazy....


arfoll:
@tingleby can you check 93cf9d24c9352cf8172467b26017c1a3e8ac14b5 is ok?

I like the gpio.c sample but I think we should call it mraa-gpio.c -> mraa-gpio and install it to ${bindir} by default. Any objections?

tingleby:
Yeah looks roughly Ok. I'm probably going probably going to introduce a flag for overriding the Arch setting.

`Acked-by: Thomas Ingleby <thomas.c.ingleby@intel.com>`

arfoll:
Thanks. Merged (finally). I renamed gpio.c to mraa-gpio.c (and the accompanying cmake target etc...)

PR #180, opened 2015-03-29
Arm beaglebone

MistreanuIonutCosmin:


arfoll:
Looks awesome thanks Michael - I'll rebase/clang-format these and merge it in.

FYI the SW CS stuff is still stuck in my backlog, I need a few hours to decide what I really want to call these functions.

michael-ring:
Great!

To make your backlog a little longer, I will soon be able to support BananaPi/Pro (95%) and ODroid C1 (80%). 

I totally forgot about the Beaglebone Port, mostly I played with Mainline Kernels on Banana and Beaglebone as Fedora 22 will bring much better support for a number of Microcontroller Boards. 

Having Fedora on Intel Edison (some time in the future) would then be paradise for me.... especially with mraa available directly through Fedora Distribution, that would be the icing on the cake....

michael-ring:
Not that I know, I guess nothing official will happen until the intel patches find their way in the mainline kernel. But doing something similar as ubilinux should be doable, perhaps a nice project in early summer for me.

arfoll:
Ok i've rebased/cleaned and put in a branch - can someone confirm this works as expected - https://github.com/intel-iot-devkit/mraa/tree/bbb

arfoll:
Merged with d9d55e42bc956d3a1afa8bbc008312ebdd7742d6

It would be good if you could document how to use BBB a little in a markdown file in docs/

PR #184, opened 2015-04-05
Initial Java support added

izard:
Added SWIG target for Java and a simplest example. To build, set JAVA_HOME to the java root path and cmake -D BUILDSWIGJAVA=ON


arfoll:
You need to sign off *all* your commits - not just the first one :)

PR #185, opened 2015-04-07
First MRAA test cases suite

cosscat:
Hello,

These are the first test cases suite for mraa. I admit i am little bit nervous and have my fingers crossed, but i hope everything is fine. 
I also created a readme file that explains the next phases that i will do.

Thank you,

Costin


rwaldron:
This is awesome. I'm inspired to write a JS version once this lands.

arfoll:
@rwaldron that would rather awesome, my plan is to have this run on `make check` or something @cosscat will be running this automatically at some point on every commit (and he'll have to provide pictures!)

rwaldron:
> my plan is to have this run on make check

Thanks, I will make sure that the JS version runs with whatever mechanism you implement for this initial set.

arfoll:
I've committed part of this and added bits so it runs with `make check`. I could run it on an edison + arduino and assume it'll run nicely on any arduino based board. @cosscat currently only general + gpio is pushed, the others need more work before they get included. It also runs correctly on my dev machine (obviously most tests get skipped :))

Generally the modules need a bit of work - unittest.setUp() needs to be used instead of the init hack you have, and there is no need to use class level vars (self.) in all these methods, I've cleaned up the Gpio one but the others need to same treatment. I think the tests should be skipped if platform is not valid, see the `@u.skipIf()` tag you can add on functions/classes. For now I'll close this PR and let you resubmit the other modules when fixed.

PR #186, opened 2015-04-07
Fixed Pin assignment for Banana Pi and wrong names for SPI devices

michael-ring:
I found two small issues, both had nothing to do with your work but were errors on my side.

Code is now tested on mainline kernel and 3.4.x kernel (which is distributed by lemaker.org)

Will add docu page tomorrow

### DIFF
--- a/docs/banana_pi.md
+++ b/docs/banana_pi.md
@@ -1,0 +1,76 @@
+Banana Pi/Pro    {#bananapi}
+============
+
+The Banana Pi/Pro is a clone of the well known Raspberry Pi.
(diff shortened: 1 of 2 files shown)


arfoll:
Looks good, however I need a sign-off line before I can merge anything. Also small note you don't need to prefix banana.c with src/arm/ there's only one file we can guess what it is :) Run the following to modify commit msg and add your signoff:

```
git commit --amend --signoff
git push --force
```

arfoll:
Great, thanks for the docs - do you have time to do the same for BBB? This is all now merged in master.

PR #195, opened 2015-04-27
Added Raspberry Pi 2 board detection for Element 14 boards

anonymouse64:
When using MRAA on a Raspberry Pi 2 made in China from Element 14, the board is incorrectly identified as a Raspberry Pi Model B rev 1. 
The revision number for this board is a21041, while on other Raspberry Pi 2 boards, the revision number is a01041 (which is correctly identified).


michael-ring:
You have duplicated the whole else if for the new model, can you please change you fix so that it looks more like this:
                } else if (strstr(line, "a01041") || strstr(line, "a21041") ) {
as the two are basically the same they should share the same if, this makes it easier to see that they are identical.

arfoll:
I'm amazed, a few hours before you sent this, I got a patch from Paul Barrette doing basically the same thing but the way @michael-ring suggested... I've applied it with f6e87f63e05c1c3e79c2b7bbeea7a77e80d86fcb.

Thanks alot for your continuing efforts!

FYI - if you work on a branch rather than on master you'll be able to send nicer patches without the merges. Just regularly update your master to match intel-iot-devkit/mraa master and then rebase your branch onto master, that way the PRs sent will be cleaner.

anonymouse64:
Okay, thanks for the heads-up.

Willjobs94:
I'm having this problem, mraa think I'm on a raspberry pi 1 but I'm on a Pi 2

arfoll:
Could you debug around mraa_raspberry_pi and see where it's going wrong? What's the output of `cat /proc/cpuinfo`? I'd recommend just setting gdb inside that function and having a look at what platform ends up being detected. Likely fix is trivial but as always requres a bit of hunting :/

Willjobs94:
cat /proc/cpuinfo
> processor	: 0
> model name	: ARMv7 Processor rev 4 (v7l)
> BogoMIPS	: 38.40
> Features	: half thumb fastmult vfp edsp neon vfpv3 tls vfpv4 idiva idivt vfpd32 lpae evtstrm crc32 
> CPU implementer	: 0x41
> CPU architecture: 7
> CPU variant	: 0x0
> CPU part	: 0xd03
> CPU revision	: 4
> 
> processor	: 1
> model name	: ARMv7 Processor rev 4 (v7l)
> BogoMIPS	: 38.40
> Features	: half thumb fastmult vfp edsp neon vfpv3 tls vfpv4 idiva idivt vfpd32 lpae evtstrm crc32 
> CPU implementer	: 0x41
> CPU architecture: 7
> CPU variant	: 0x0
> CPU part	: 0xd03
> CPU revision	: 4
> 
> processor	: 2
> model name	: ARMv7 Processor rev 4 (v7l)
> BogoMIPS	: 38.40
> Features	: half thumb fastmult vfp edsp neon vfpv3 tls vfpv4 idiva idivt vfpd32 lpae evtstrm crc32 
> CPU implementer	: 0x41
> CPU architecture: 7
> CPU variant	: 0x0
> CPU part	: 0xd03
> CPU revision	: 4
> 
> processor	: 3
> model name	: ARMv7 Processor rev 4 (v7l)
> BogoMIPS	: 38.40
> Features	: half thumb fastmult vfp edsp neon vfpv3 tls vfpv4 idiva idivt vfpd32 lpae evtstrm crc32 
> CPU implementer	: 0x41
> CPU architecture: 7
> CPU variant	: 0x0
> CPU part	: 0xd03
> CPU revision	: 4
> 
> Hardware	: BCM2709
> Revision	: a02082
> Serial		: 000000007326637f

how can I do the workaround of setting gbd(and what is) ?

arfoll:
ok so the issue is that your revision on your pi is newer than what we've seen. I should have fixed it in master, just rebuild from src following the instructions in docs/building.md and it should work. If it doesn't please file a a new bug.

Willjobs94:
When i finished the build and move the files to node_modules, my script does not find the mraa package.
I'm I doing something wrong ?

arfoll:
well do you end up with a mraa.node file and you placed that in /usr/lib/node_modules/mraa/ alongside a package.json file? That should work as long as your NODE env path is correct

Willjobs94:
Not i didn't

arfoll:
so what was the output of `make install`?

Willjobs94:
screenshot 2016-03-22 13 14 57

arfoll:
Add `-DCMAKE_INSTALL_PREFIX:PATH=/usr`, your node modules are all going to
/usr/local not /usr. Please use dpaste.com instead of posting images ;-)

Willjobs94:
This is a flag for `sudo make install` ? Is so, is not working when I run `sudo make install -DCMAKE_INSTALL_PREFIX:PATH=/usr ` 
[Error](http://dpaste.com/02CZ5EG)
And why not pastebin.com ?

arfoll:
check building.md again ;-) You use that flag against configure like `cmake -DCMAKE_INSTALL_PREFIX:PATH=/usr ..`. To change where make install points you have to use `DESTIDR=` but that doesn't work all the time so best reconfigure.

Not pastebin.com because there's horrible adverts there xD ?

arfoll:
That prefix needs to be added after cmake is run (just after you create the
build dir). Please read building.md  thoroughly. There are no options you
need/should give to make.

PR #196, opened 2015-04-28
java: capitalized file names of examples

jkramarz:
Without this change, on Linux compilation would be unsuccessful:
```
example.java:25: error: class Example is public, should be declared in a file named Example.java
public class Example {
```


jkramarz:
Duplicated by #197

PR #199, opened 2015-04-29
Pull MRAA UART enhancements

Jon-ICS:
These commits add some additional capabilities to the UART support in MRAA.

Specifically, the ability to open, close, read and write to a TTY device.

Additionally, a method to check the device (with time out) to see if data is available to be read.

The following MRAA functions are added (along with their .hpp counterparts):

mraa_result_t 
mraa_uart_open_dev(mraa_uart_context dev, speed_t baud)

mraa_result_t 
mraa_uart_close_dev(mraa_uart_context dev)

int 
mraa_uart_read(mraa_uart_context dev, char *buf, size_t len)

int 
mraa_uart_write(mraa_uart_context dev, char *buf, size_t len)

mraa_boolean_t 
mraa_uart_data_available(mraa_uart_context dev, unsigned int millis)

### DIFF
--- a/api/mraa/uart.h
+++ b/api/mraa/uart.h
@@ -1,5 +1,6 @@
 /*
  * Author: Thomas Ingleby <thomas.c.ingleby@intel.com>
+ * Contributions: Jon Trulson <jtrulson@ics.com>
  * Copyright (c) 2014 Intel Corporation.
  *
  * Permission is hereby granted, free of charge, to any person obtaining
@@ -41,6 +42,15 @@ extern "C" {
 
 #include <stdio.h>
 #include <fcntl.h>
+#include <stdlib.h>
+#include <unistd.h>
+#include <string.h>
+#include <errno.h>
+#include <termios.h>
+#include <sys/time.h>
+#include <sys/select.h>
+#include <sys/types.h>
+#include <sys/stat.h>
 
 #include "common.h"
 
@@ -62,6 +72,26 @@ mraa_uart_context mraa_uart_init(int uart);
  * @return char pointer of device path
  */
 char* mraa_uart_get_dev_path(mraa_uart_context dev);
+
+/**
+ * Open the TTY device associated with a UART context, and set up the
+ * terminal modes and baud rate.  The TTY is setup for a 'raw'
+ * mode. 81N, no echo or special character handling, such as flow
+ * control or line editing semantics.
+ *
+ * @param dev uart context
+ * @param baud desired baud rate
+ * @return mraa_result_t
+ */
+mraa_result_t mraa_uart_open_dev(mraa_uart_context dev, speed_t baud);
+
+/**
+ * Close a device previously opened with mraa_uart_open_dev().
+ *
+ * @param dev uart context
+ * @return mraa_result_t
+ */
+mraa_result_t mraa_uart_close_dev(mraa_uart_context dev);
 
 #ifdef __cplusplus
 }
--- a/api/mraa/uart.h
+++ b/api/mraa/uart.h
@@ -83,7 +83,7 @@ char* mraa_uart_get_dev_path(mraa_uart_context dev);
  * @param baud desired baud rate
  * @return mraa_result_t
  */
-mraa_result_t mraa_uart_open_dev(mraa_uart_context dev, speed_t baud);
+mraa_result_t mraa_uart_open_dev(mraa_uart_context dev, unsigned int baud);
 
 /**
  * Close a device previously opened with mraa_uart_open_dev().
--- a/api/mraa/uart.h
+++ b/api/mraa/uart.h
@@ -113,6 +113,15 @@ int mraa_uart_read(mraa_uart_context dev, char *buf, size_t len);
  */
 int mraa_uart_write(mraa_uart_context dev, char *buf, size_t len);
 
+/**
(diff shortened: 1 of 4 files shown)


tingleby:
This is certainty something that was on the potential TODO list.
One key thing that should be seen with with UART API features is portability beyond Linux.
So for your baud call I think either a enum or int should be use.

Jon-ICS:
Are there other operating systems besides Linux? :)
/kidding

I see your point.  How about I make it an unsigned int, and internally map it to the correct speed_t items internally.  That way in the future, custom baud rates might be supported as well.

tingleby:
Non-Linux, a dark place we shall not speak of /s :)
Yes The unsigned int would be best, as some hardware can support very obscure baud rates.

arfoll:
This is really good, thanks for driving this @Jon-ICS . I'd rather not have open/close functions and use the standard way other mraa bits work. How about making it like this:

```
mraa_uart_init(int uart); <- opens uart device that's supported
mraa_uart_init_raw(int uart); <- opens any tty device
mraa_uart_set_baudrate(int baudrate); <- sets baudrate
mraa_uart_set_mode(bytesize, parity, stopbits); <- with some enums to match pyserial
mraa_uart_set_flowcontrol(xonxoff, rtscts, dsrdtr); <- bools I think right?
mraa_uart_set_timeout(read, write, char); <- seems to be in seconds in pyserial, sounds reasonable
```
I've grouped stuff together a bit otherwise it gets silly long. In C++ they could all be put as optional params matching pyserial API - http://pyserial.sourceforge.net/pyserial_api.html#classes

The read/write functions look good as they are obviously they should respect the timeout values.

What do you guys think?

Jon-ICS:
On 04/30/2015 06:40 AM, Brendan Le Foll wrote:
>
> This is really good, thanks for driving this @Jon-ICS 
> <https://github.com/Jon-ICS> .
>

Sure!

> I'd rather not have open/close functions and use the standard way 
> other mraa bits work.
>

Well, you do really need a close function so that the associated file 
descriptor (fd) can be released properly.  There are other MRAA close 
functions, such as _gpio_close(), aio_close(), so a _uart_close() 
shouldn't be an API breaker, and as I said, you do need to close that FD 
when done.

Also, I like the 'separation of powers' - mraa_uart_init() is just 
responsible for creating the context and setting up the pins.  _open if 
for actually opening the device and doing initial setup.  In some cases 
that may be all that someone wants (mraa_uart_init()), allowing them to 
handle the port on their own.

> How about making it like this:
>
> |mraa_uart_init(int uart); <- opens uart device that's supported
> mraa_uart_init_raw(int uart); <- opens any tty device
> |

I am not even sure how an init_raw would work... I am not an expert in 
MRAA internals, but that sounds like something you guys should do :)

> |mraa_uart_set_baudrate(int baudrate); <- sets baudrate|

I could add a separate function to set the baud rate if desired.

> |
> mraa_uart_set_mode(bytesize, parity, stopbits); <- with some enums to match pyserial
> mraa_uart_set_flowcontrol(xonxoff, rtscts, dsrdtr); <- bools I think right?
> mraa_uart_set_timeout(read, write, char); <- seems to be in seconds in pyserial, sounds reasonable
> |
>
> I've grouped stuff together a bit otherwise it gets silly long. In C++ 
> they could all be put as optional params matching pyserial API - 
> http://pyserial.sourceforge.net/pyserial_api.html#classes
>
> The read/write functions look good as they are obviously they should 
> respect the timeout values.
>
> What do you guys think?
>

The functionality I added was designed around supporting the UPM drivers 
I've written (and others yet to be written) - it was not designed to be 
a full-fledged serial facility like pyserial, nor did I have any desire 
to emulate another language's serial implementation :)

MRAA already provides the necessary interfaces 
(mraa_uart_get_dev_path()) that would allow the use of the full 
capabilities of a 'full' serial port.  I believe nodejs has that as well 
(https://github.com/voodootikigod/node-serialport).

Hardware flow control is never used in reality on these sensors, and I 
have not seen one yet that supported software flow control either. These 
functions could be added on if needed (maybe SW flow control), but...

I could add a function to support 7b, odd/even parity and 1 or 2 stop 
bits I suppose, but I haven't seen that in the wild either. The serial 
implementation on these devices are designed to be as simple as 
possible, and I haven't even seen a 7bit, or odd/even parity or 
2-bit-stop device in like 20 years :)

As for the delays/timeouts  - I would very much prefer:

1. A seperate function for testing the availablity to read (and write if 
needed).  I've only ever needed to test for readability, but I can see 
how it might be useful to test for writability as well, but I'd rather 
keep them as separate functions, perhaps something like:

mraa_uart_read_possible(unsigned int millis)
mraa_uart_write_possible(unsigned int millis)

I believe read/write should always be blocking.  Non-blocking I/O is not 
the same thing as what I've implemented here.  True non-blocking i/o is 
a different animal entirely and requires careful design of the code 
making use of it.

and

2. I would rather have millisecond granularity than second granularity.  
As these will block, I'd like the option to block only the minimum time 
required, which is rarely more than 100ms, and in some packet timing 
scenarios, 10ms.  Seconds can of course be done by specifying the 
appropriate time in millis if that is really needed :)

arfoll:
Closing as this now on uart branch and being wored on. Will be on master soon. Thanks!

PR #201, opened 2015-05-05
fixed Raspberry Pi model 2 pin mapping docs

jkramarz:
It weren't rendered correctly by Github in previous version (see rich diff); also the header was missing.


arfoll:
Thanks alot! Merged.

PR #203, opened 2015-05-08
de3815.c: Fix the I2C bus_id incorrect issue

yongli3:
Signed-off-by: Yong Li <yong.b.li@intel.com>

I found the I2C bus on DE3815 should be 8 and 9, this change fixes the incorrect bus id


tingleby:
I'm afraid its going to be more intelligent then that.
As demonstrated by the original bug. It depends on how the I2C-dev are loaded. Changing the bus number will break it for other OSed

tingleby:
And by OSes I mean distros

yongli3:
Thanks Tom! You can right, we need to consider the kernel version and I2C modules loading order. Let me test more

PR #214, opened 2015-06-01
Pull various uart fixes

Jon-ICS:
- uart: undo changes to write() making use of std::string
- uart: revert changes to use a std::string in read()
- uart: fix missing inversion that will clear all other flags
- uart: set fd to -1 after close to prevent accidental re-use

### DIFF
--- a/api/mraa/uart.hpp
+++ b/api/mraa/uart.hpp
@@ -99,19 +99,12 @@ class Uart
      *
      * @param data buffer pointer
      * @param length maximum size of buffer
-     * @return string of data
+     * @return number of bytes read, or -1 on error
      */
-    std::string
-    read(int length)
+    int
+    read(char* data, int length)
     {
-        char* data = (char*) malloc(sizeof(char) * length);
-        int v = mraa_uart_read(m_uart, data, (size_t) length);
-        char* out = (char*) malloc(sizeof(char) * v);
-        strncpy(out, data, v);
-        std::string ret(out);
-        free(data);
-        free(out);
-        return ret;
+        return mraa_uart_read(m_uart, data, length);
     }
 
     /**
--- a/api/mraa/uart.hpp
+++ b/api/mraa/uart.hpp
@@ -111,15 +111,13 @@ class Uart
      * Write bytes in buffer to a device
      *
      * @param data buffer pointer
-     * @param length maximum size of buffer
+     * @param length length of buffer
      * @return the number of bytes written, or -1 if an error occurred
      */
     int
-    write(std::string data)
+    write(char* data, int length)
     {
-        char *d = new char[data.length() + 1];
-        std::strcpy(d, data.c_str());
-        return mraa_uart_write(m_uart, d, (data.length() + 1));
+        return mraa_uart_write(m_uart, data, length);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


arfoll:
Thanks alot Jon, really appreciate your help. I've handled the binary data a little different and moved the string functions to 'Str' variants because I do think they're useful but as you say they don't fit every case. I don't think b98cea2 is necessary but 049d189 is a good catch so I pulled that one in.

Jon-ICS:
Ah, so you just made them overloaded methods then?  Awesome, thanks!  Sorry about the fd = -1, did not realize you freed the struct right after. :) But your Str methods will still need to be fixed - using memcpy/memmove rather than str[n]cpy() at least.

arfoll:
Please look at the newest uart.hpp you'll see no strcopy/memcpy there :)

Jon-ICS:
Just pulled, much better :)  Thanks!
-jon

On 06/01/2015 10:21 AM, Brendan Le Foll wrote:
>
> Please look at the newest uart.hpp you'll see no strcopy/memcpy there :)
>
> —
> Reply to this email directly or view it on GitHub 
> <https://github.com/intel-iot-devkit/mraa/pull/214#issuecomment-107620177>.
>

-- 
Jon Trulson
Integrated Computer Solutions
www.ics.com

Jon-ICS:
Ok, so I got the SM130 over the weekend, and had already received the 
RFID eval breakout shield from sparkfun (required in order to connect 
and use it).

Today was spent assembling it and figuring out the status of the driver.

The status is... confusing.

There is no example for this driver, and looking at it in more detail, 
it is not even complete.  It looks like it was started, then abandoned 
shortly thereafter, yet still merged into UPM :)

I was unable to get anything using I2C (on either edison or G2).  As 
it's interfaced via the eval shield, I did not have the option to scope it.

Looking through the data sheet, it appears this device only supports a 
UART interface, unless you want to contact their support department, 
obtain an encrypted firmware blob, and then use a windows utility (and 
unknown connection apparatus) to reflash the firmware to a version that 
allows I2C comms.  This dates to 2008, so I am curious why their current 
offerings don't already support this, but then again, it appears their 
website hasn't been updated in years).

Using the edison, I was able to finally get UART comms working well 
enough to find out the firmware version (and therefore discover there is 
no I2C support).  The eval RFID shield assumes arduino and software 
serial, so I had to rework the UART selection pads to get it to use the 
HW UART.  Took a couple of times as those pads are tiny and I kept 
shorting them incorrectly.  When you do that, the edison/G2 won't power 
up :)

So, what would you like me to do?  I see 3 options:

 1. throw out the current driver and re-do as a UART driver (with
    instructions in the header file on proper setup of the eval shield
    you connect it to).  At least on edison, this is probably the
    easiest path. G2 has a habit of throwing away UART data if it's too
    large (it doesn't seem to buffer more than a handful of bytes).
 2. look into what is required to update the firmware to an i2c version,
    do so if possible, and if successful, implement an i2c driver (also
    mostly throwing out the current code anyway).
 3. just remove this 'half-driver' from UPM and point people towards
    pn532 :)


What do you guys think?

-- 
Jon Trulson
Integrated Computer Solutions
www.ics.com

Issue #178, opened 2015-03-25
python SPI: lsbmode() accepts False but not True


cosscat:
Hello,

On Galileo v1, the lsbmode(True) returns errcode 7:

>>> spi=m.Spi(0)
>>> spi.lsbmode(False)
0
>>> spi.lsbmode(True)
7
>>>
>>> m.getVersion()
'v0.6.1'

Costin

arfoll:
galileo g1 does not seem to support lsbmode LSB_FIRST. Thus the ioctl's fail at a kernel level returning MRAA_ERROR_INVALID_RESOURCE (ideally on fail it should return MRAA_ERROR_FEATURE_NOT_SUPPORTED) we should add a _replace() hook for this on g1.

Issue #22, opened 2014-10-23
[Edison] mraa_spi_lsbmode Does not have any affect


versamodule:
using mraa_spi_lsbmode(spi,0) or mraa_spi_lsbmode(spi,1) does not have any affect on how data is sent. 

also, the function for it could be trimmed up a bit by getting rid of the first if statement. to this:

mraa_result_t
mraa_spi_lsbmode(mraa_spi_context dev, mraa_boolean_t lsb)
{
    uint8_t lsb_mode = (uint8_t) lsb
...

Yoda12999:
Is there a possibility that the SPI device on the Edison does not support LSB mode? I have looked at the documentation for the Atom processor and it says nothing about this, but this is a rather odd feature.

alext-mkrs:
While preparing a fix for issue #178 I've came across this issue too. I'm going to look more into this to see if I can find any additional clues, however as a data point - below is what I see in the `dmesg` output when trying to set lsbmode to true (the log has several attempts listed, it's one line per one lsbmode function invocation):
```
[  697.786230] spidev spi5.1: setup: unsupported mode bits 8
[  723.823990] spidev spi5.1: setup: unsupported mode bits 8
[  737.880541] spidev spi5.1: setup: unsupported mode bits 8
[  761.673874] spidev spi5.1: setup: unsupported mode bits 8
[ 2488.659357] spidev spi5.1: setup: unsupported mode bits 8
```

alext-mkrs:
After looking into the driver and Linux SPI sources I would say it is indeed the case that Edison's SPI controller we use in MRAA doesn't support LSB_FIRST mode, just as that log message I pasted above confirms.

@arfoll, shouldn't we implement the same _replace hook function for Edison as for the Galileo Gen1 in #178?

arfoll:
Yes, the same _replace hook would probably be a good idea. Are you willing to send a patch for this as well?

alext-mkrs:
Yep, will do

PR #223, opened 2015-06-14
Adding spi_lsbmode_replace() for Galileo Gen1 to return proper error code

alext-mkrs:
This is for issue #178, I've implemented the _replace hook (that required some prerequisites to be in place first - all are in separate commits to have a logical separation) to return proper error code.

I haven't tested it on namely Galileo (I don't have a Linux toolchain for Iot DevKit image and couldn't find neither sources nor a prebuilt one - if you have any pointers, I'd be grateful for a suggestion).

However I've tested it on Edison (by adding identical code to Edison't platform file) - worked as expected. By the way Edison seems to also have problems with LSB mode (already captured in issue #22).


arfoll:
Awesome, thanks alot! All Merged

PR #239, opened 2015-07-15
java: fixed missing flags for cross-compiling java bindings

Propanu:

### DIFF
--- a/src/java/CMakeLists.txt
+++ b/src/java/CMakeLists.txt
@@ -10,7 +10,7 @@ include_directories (
 set_source_files_properties (mraajava.i PROPERTIES SWIG_FLAGS ";-package;mraa;-I${CMAKE_BINARY_DIR}/src")
 set_source_files_properties (mraajava.i PROPERTIES CPLUSPLUS ON)
 
-set (CMAKE_CXX_FLAGS "-fpermissive")
+set (CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fpermissive")
 
 set (JAVAC $ENV{JAVA_HOME}/bin/javac)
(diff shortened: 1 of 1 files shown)


arfoll:
Can you clarify what error you where getting without it? I've compiled the java bindings a fair bit and never come accross this issue... Which gcc/java version are you getting this on?

Propanu:
Typical 32-bit target compiled on 64-bit host kind of errors:

`/usr/bin/ld: i386 architecture of input file 'CMakeFiles/mraajava.dir/__/mraa.c.o' is incompatible with i386:x86-64 output`

If you don't add the variable in there, your global CXX flags get lost so the objects get built for the host architecture. Look at some of the other cmake files, they actually preserve the flags when they need to be set again. GCC 4.8.2 and java 1.7 but not relevant to this issue.

arfoll:
That's because you're not cross compiling correctly, use the oe-cross-sdk toolchain file for example or create your own toolchain file. You can't 'cross' compile by simply adding a few `-m32`, doing what you describe in upm/building.md:
`
-DCMAKE_CXX_FLAGS:STRING=-m32 -march=i586
-DCMAKE_C_FLAGS:STRING=-m32 -march=i586
`
This is just wrong, you're just telling your host compiler to generate 32bit code on a 64bit host - expect host contamination! If you really want to disable -fpermissive then doing it with -DCMAKE_C_FLAGS I won't disable these warnings by default.

arfoll:
Spoke too soon, didn't realise -fpermissive was actually required to compile the wrapper code *fail*. I merged with 296172d38550a372c00dc4cef569acbaee15b3d7 and modified the commit message to match. Cheers!

PR #273, opened 2015-09-05
Fix for PWM on Beaglebone black with a kernel > 3.8.13

SyrianSpock:
Tested on hardware (Beaglebone black Rev B6)

### DIFF
--- a/src/pwm/pwm.c
+++ b/src/pwm/pwm.c
@@ -38,6 +38,9 @@ mraa_pwm_setup_duty_fp(mraa_pwm_context dev)
 {
     char bu[MAX_SIZE];
     snprintf(bu, MAX_SIZE, "/sys/class/pwm/pwmchip%d/pwm%d/duty_cycle", dev->chipid, dev->pin);
+    if (!mraa_file_exist(bu)) {
+        snprintf(bu, MAX_SIZE, "/sys/class/pwm/pwm%d/duty_ns", dev->pin);
+    }
 
     dev->duty_fp = open(bu, O_RDWR);
     if (dev->duty_fp == -1) {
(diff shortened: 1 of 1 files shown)


arfoll:
Thanks for the contribution! Is this a debian thing or global? I've never seen such a thing and edison for example has a 3.10 kernel? I'd like to understand if this is a beaglebone thing or general?

SyrianSpock:
I have only seen this on the Beaglebone black with Debian and Ubuntu.
I don't know if the Beaglebone is the only board concerned or if it's an Linux on ARM thingy, sorry I don't have an answer.

arfoll:
Ok I unfortunately don't have a BBB, @tingleby do you think you could take a look?

arfoll:
@cosscat is this something you could help with if you have a board?

alext-mkrs:
I don't have BBB either, but I've found this nice description: https://github.com/szmoore/MCTX3420/wiki/Software:-PWM-Control and looks like the standard sysfs location is actually just a secondary mean to use PWMs on BBB and the one that doesn't follow the kernel documentation fully (dropping the pwmchip piece).

alext-mkrs:
I believe this should rather be reworked using `_replace` approach, as this is specific to one board we support. @SyrianSpock, do you think you'd want to do that?

SyrianSpock:
@alext-mkrs Sure, I'll do it ASAP.

alext-mkrs:
Thanks and no rush :smile: I just believe this is a better way to go and looking at the code and @arfoll's comments I think it will increase its chances for merging.

alext-mkrs:
So @SyrianSpock, would you be willing to modify this one as requested?

PR #274, opened 2015-09-08
Edison: Don't initiate SPI pins when initializing MRAA

jonasconnode:
My problem is that I have a daemon running in the background that uses the
SPI interface on an Edison/Arduino board. A program using MRAA would disable
the SPI bus when started, so this commit makes MRAA a bit more considerate.

Basic GPIO and SPI functionality has been tested (with Python) after this change.

---

Calling mraa_intel_edison_misc_spi() is not really needed, because the
GPIO 10..13 pins will be configured correctly when the user actually
inits them as GPIO pins. When using these pins for GPIO,
mraa_setup_mux_mapped() and mraa_intel_edison_gpio_init_post() will do
all this work based on the pin map and pinmodes for the Edison Arduino
board.

On the contrary, this function would break any user of the SPI bus
that is already running.

### DIFF
--- a/src/x86/intel_edison_fab_c.c
+++ b/src/x86/intel_edison_fab_c.c
@@ -247,46 +247,6 @@ mraa_intel_edison_i2c_init_pre(unsigned int bus)
     return MRAA_SUCCESS;
 }
 
-static mraa_result_t
-mraa_intel_edison_misc_spi()
-{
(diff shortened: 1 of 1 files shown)


arfoll:
Thanks for the contribution, I'll make sure we test this thoroughly and merge in the next few days!

arfoll:
Thanks alot, merged!

PR #276, opened 2015-09-09
Added MinnowBoard Compatible support

steelee:
This change allows libmraa to function with the new ADI Minnowboards. Additionally,
the Minnowboard files were changed to better incorporate new boards, hopefully making
future board additions easier. This change has been tested on the Minnowboard MAX and
the ADI Minnowboard Turbot.

Signed-off-by: Evan Steele <evan.steele@intel.com>


arfoll:
Thanks - mostly merged with some edits, I don't want to change the typenames since compatible boards are - for now - entirely identical. Can revisit when this is no longer the case.

PR #287, opened 2015-09-18
intel_minnow_byt_compatible: fix memory leak

matt-auld:
mraa_adv_func_t was allocated twice for the adv_func field


arfoll:
Merged with 434f45413976028164fa54cbe3e8cd167016fe76. Thanks alot!

PR #291, opened 2015-09-21
mraa: updated travis file to use swig3 and build node.js bindings too

Propanu:
BUILDSWIGPYTHON and BUILDSWIGNODE are on by default so this should work as is.


arfoll:
Thanks alot Tudor this is great!

PR #297, opened 2015-10-01
Add nodejs 4.1.1 support

xbolshe:


arfoll:
This is really great thank you, however two small things (I can do the first for you if you prefer but the second has to be you :)):

1. The way you set c++11 doesn't work (at least for me) and it does look like it's required on nodejs 4.x, I set it correctly with this in src/javascript/CMakeLists.txt:
```
if (${V8_VERSION_MAJOR} GREATER 3)
  # Yup node 0.12.x major version is '3'...
  message (INFO " - requiring C++11 since using node.js ${V8_VERSION_MAJOR} > 3")
  set_property (TARGET mraajs PROPERTY CXX_STANDARD 11)
  set_property (TARGET mraajs PROPERTY CXX_STANDARD_REQUIRED ON)
endif ()
```
2. You can't sign off with an anonymous name/nickname like 'xbolshe', please use your real name to signify acceptance of https://github.com/intel-iot-devkit/mraa/blob/master/docs/contributing.md

xbolshe:
Thank you for hints. I will correct it. As I see nodejs 3.x and 2.1.x also require c++11, because they use Google's v8 engine version 4.x.

xbolshe:
About "The way you set c++11 doesn't work (at least for me)": what a compiler version do you use? I use GCC 4.9.1. -std=c++11 requires GCC 4.7 or above.

arfoll:
That was on GCC 5.2. But it might have been my cmake cache not working
correctly

haroldiedema:
@xbolshe, I've tried building mraa with your fork, but I'm getting strange results :/

Output of cmake (works): https://gist.github.com/haroldiedema/c718875d1edf99fa3908
Output of make (fails): https://gist.github.com/haroldiedema/c7958e335b19b0b30fa0

Building on Edison running latest ubilinux.
As far as I know, everything is up-to-date.

gcc version 4.7.2 (Debian 4.7.2-5)
Node 4.1.1
Swig 3.0.5

Is this a problem on my end?

Propanu:
I agree with @arfoll, set_property is the right way to do it. Travis CI for instance has gcc 4.6 and thus will only take --std=c++0x.

xbolshe:
https://github.com/Propanu gcc 4.6 has -std=c1x, but it is a partitial c++11 implementation. And it is not recommended to use it. I have read a documentation (https://cmake.org/cmake/help/v3.1/prop_tgt/CXX_STANDARD.html) and tested set_property, and also agree with https://github.com/arfoll.

xbolshe:
@haroldiedema In your case c++11 was not turned on. I will update a code.

xbolshe:
Tested on Galileo Gen1:
mraa_test_1

haroldiedema:
@xbolshe 

When I run make, I now get an INFO line saying "requiring C++11 since using node.js with V8 4.5.103.33". However, when I run make, I still get similar errors.

After installing C++11 (using apt-get), running cmake and make again, nothing appears to change.

Here are both of the outputs of make:

Without C++11: http://pastebin.com/0LqrApFz
With C++11: http://pastebin.com/PVqUvED6

Both outputs appear to be identical. Do I need to do something to enforce compiling with C++11 on my end?

Apologies for asking silly questions, my knowledge in C++ is very limited :(

arfoll:
Paste the output of cmake in a clean build directory. The cmake feature
cxx11 should be set to required rather than optional then it'll fail the
builds. I'll give this a good run on Monday (or maybe before if i get the
time) and squash your patches down before merging.

haroldiedema:
Oh right, apologies @arfoll.

Here's the CMAKE output with C++11 installed: http://pastebin.com/MTdD3qM0
And here it is without it: http://pastebin.com/8jqgfBz8

xbolshe:
@haroldiedema C++11 is not a software package in this context. You use "gcc 4.7.2" in both cases, which support c++11 option. So, outputs are the same. In case of gcc 4.6 an error message will be shown: https://travis-ci.org/intel-iot-devkit/mraa/jobs/83415933

xbolshe:
mraa_test_2

haroldiedema:
Well... I don't know what else to say, but it still doesn't work.

I'm using the latest node binaries and header files (downloaded from nodejs.org)
From what I understand of the errors I'm getting, is that whenever the javascript wrapper gets compiled, it's not compatible with the v8 header file in `/usr/include/node/v8.h`.

arfoll:
@xbolshe Rebased & merged with 3059c06fa05c4a5536ec82d00bd991891956704d. Please avoid adding samples to unrelated commits and name your commits individually & appropriately. IMHO Samples for gen1 blink can go with the generic blink one, the uart one can replace the current very poor uart.js rather than being a new file. There are already 2 i2c samples so I'd rather not take it unless there's a good reason.

@haroldiedema please open a seperate bug if you're still having issues. Here are my general ideas:
1. GCC 4.7.x is probably too old, it's not full C++11 compliant (see https://gcc.gnu.org/gcc-4.7/cxx0x_status.html). You can't use the latest V8 and then a compiler that's years old and expect no pain!
2. Binaries from nodejs.org are actually quite bad since they miss out some of the headers, they're really badly packaged, I've complained to them before but they missed the point, compile it from source. Check SWIG_V8_VERSION is being set correctly in the mraajs wrapper file.
3. Ubilinux is a massive hack, we've seen a bunch of issues being reported with it so I simply don't support it. I'll compile a node 4.x on my debian box, if it works there I'm claiming it yet another ubilinux issue. I'd highly recommend moving off it.

Issue #268, opened 2015-09-02
Edison having issues with gpio ISR in python


cosscat:
Hi. how to reproduce:
in python:

>>> 
>>> import mraa as m

>>> def func():
...   print "hello"
... 
>>>io=m.Gpio(3)                                                                                                                                                                                                              
>>> io.dir(m.DIR_IN) # though i guess this is not necessary right?
0
>>> io.isr(m.EDGE_FALLING, func, 6)
0

>>> io.isrExit() # this step blocks the serial console. I can't do anything with it and to regain control board reset is needed

alext-mkrs:
According to my findings, this only happens when you run Python interactively (both 2 and 3) and the reason is a deadlock between `pthread_join()` in `isrExit()` (after we've sent an ISR thread cancellation command) and an attempt to get the Python global interpreter lock (GIL) in the code path it gets through when processing the thread cancellation.

I'm currently thinking on how to fix this properly, but if you need a solid workaround @cosscat, then just run it from the script and not interactively.

PR #304, opened 2015-10-02
Fix deadlock in Python console upon isrExit() + a couple of smaller things

alext-mkrs:
As mentioned in #268, there's a deadlock between our code that tries to execute the ISR (and obtain the GIL, more specifically) and the `pthread_join()` call in  `isrExit()` - it's getting stuck on `futex()` calls at both ends. This only happens when Python is in the console mode (running from script is okay, futexes are there, but don't deadlock).

I don't see any better way than to introduce a transient flag in `mraa_gpio_context` indicating that the ISR thread is being shut down. Even though the problem at hand is in some sense a corner case, it's anyway not good that we try to execute the ISR upon thread exit.

The chain before the patch is: `pthread_cancel()`->`poll()` returns like there's an interrupt->`read()`->`return MRAA_SUCCESS`->we're executing the ISR in `mraa_gpio_interrupt_handler()`. There seems to be no better way to distinguish `poll()` being interrupted by `pthread_cancel()` and a normal interrupt.

But I wouldn's say I like this approach (adding a flag is so-so), so I'd like to run this by you @arfoll. Any comments or suggestions are more than welcome.

The other two commits are a typo fix and some more elaborated error output in case our ISR fails (helped me during debugging and I think it would be generally useful).

### DIFF
--- a/include/mraa_internal_types.h
+++ b/include/mraa_internal_types.h
@@ -47,6 +47,7 @@ struct _gpio {
     void *isr_args; /**< args return when interupt service request triggered */
     pthread_t thread_id; /**< the isr handler thread id */
     int isr_value_fp; /**< the isr file pointer on the value */
+    mraa_boolean_t isr_thread_terminating; /**< is the isr thread being terminated? */
     mraa_boolean_t owner; /**< If this context originally exported the pin */
     mraa_result_t (*mmap_write) (mraa_gpio_context dev, int value);
     int (*mmap_read) (mraa_gpio_context dev);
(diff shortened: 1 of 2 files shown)


alext-mkrs:
@arfoll, reworked to use `mraa_boolean_t` & rebased onto current master.

arfoll:
Thanks alot, great work as usual! merged with 3c3136d43a70ad20160410fd5802cad430525cdf

alext-mkrs:
Thanks :)

Issue #299, opened 2015-10-02
iio: count only devices matching iio:deviceX when enumerating them


arfoll:
As @alext-mkrs spotted we count every symlink in /sys/bus/iio which is wrong.

alext-mkrs:
Let me start with this one, feel free to assign it to me.

PR #310, opened 2015-10-06
Fixed IIO device counting + small typo correction

alext-mkrs:
Here comes my take at #299.

A couple of notes:

1. As long as we do name matching, we arguably don't need to check if that's a symlink or not in the counter subroutine, so I removed that.
2. `IIO_DEVICE_WILDCARD` should better be co-located with (and inherited from) similar defines in `iio.c`, however as long as (a) e.g. `MAX_SIZE` (and similar ones) is redefined across various files and (b) it feels like exact header structure/contents for IIO is yet in flux, I've put it right into `mraa.c` - please let me know if you can think of a more appropriate location.


arfoll:
Pushed, thanks!

Issue #309, opened 2015-10-06
iio: allow device index to be found via device->name


arfoll:

alext-mkrs:
Let me take this one next.

alext-mkrs:
Hmm, interesting that this one is still open even though I've referenced it in the commit, @arfoll, could you please close it?

arfoll:
Maybe because it's not pushed to master? Not sure but closed.

tingleby:
Issue normally only gets actioned on with references when the commit hits master, which kind of makes sense as iio is a feature branch yet in "upsteam" mraa.

alext-mkrs:
Good point, I think that's exactly the case.

PR #317, opened 2015-10-10
Implementation of "get device index by name" (#309) + a fix for potential bug

alext-mkrs:
The main part is just what #309 is about.

The the second commit is something I've came across during the implementation - we don't normalize IIO device names we read from sysfs and at least on Edison they contain trailing LFs. That may cause unexpected results in downstream consumers of this information.

Example:
```
root@edison1:~# cat /sys/bus/iio/devices/iio\:device0/name |hexdump -C
00000000  62 63 6f 76 65 5f 61 64  63 0a                    |bcove_adc.|
0000000a
root@edison1:~#
```


arfoll:
Forgot to close, merged! Thanks Alex.

PR #319, opened 2015-10-16
iio: added event api

laykuanloon:


arfoll:
This needs a sign-off line, I'll review properly on monday

PR #327, opened 2015-10-20
Ft4222 cleanup

whbruce:
I’ve now implemented all i2c calls in the ft4222 sub-platform and fixed a gpio configuration bug.


arfoll:
Please scrap your branches, the tree is a mess! I merged I think everything I could find :) Thanks!

PR #342, opened 2015-10-28
iio: read float fix

laykuanloon:
Signed-off-by: Lay, Kuan Loon <kuan.loon.lay@intel.com>

### DIFF
--- a/src/iio/iio.c
+++ b/src/iio/iio.c
@@ -93,9 +93,9 @@ mraa_iio_get_channel_data(mraa_iio_context dev)
         }
     }
     dev->chan_num = chan_num;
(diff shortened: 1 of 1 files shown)


arfoll:
I did this a different way, using MAX_SIZE instead of 128/64 as a magic number everywhere. Should do the same. See 357b40915c3dbc3f2e910b8a7204b9e7f79cc06c

Also please don't make styling fixes with functional fixes seperate commits are much better because they are bisectable. I just used clang-format on the whole file to clean it up 3a33a6e4bce2a5cde0a2b4bfd9c9dea03d7950de

PR #344, opened 2015-10-29
gpio.c: enhanced Python-specific code

alext-mkrs:
This further enhances fix introduced by commit 0836067 by adding Python 3 support and correcting Python version hex string used in compiler directives.

The piece has become a bit longer than initially, however I think it contains a useful information for debugging misbehaving ISR under Python, so it's still worth it.


arfoll:
Merged, thanks! cd4626852d55e5f6067f01a0e1c11b86fdd8e4f4

PR #348, opened 2015-10-31
Intel Edison: Add break-out board type to Edison platform name

achilikin:
Add detected break-out board (Arduino or Miniboard) to Edison platform name


arfoll:
I'm not sure this should be done like this as it means we change for every board and then strcmp gets a lot more complex.. You can check pincount to check this so not sure of value.

In any case please docs/contributing.md commits need to be signed off.

achilikin:
Sorry, did not think about other boards... I'm using it just to distinguish logs from mini and Arduino boards. Adding support for extra boards will make things more complex as you said.

Issue #336, opened 2015-10-24
mraa not properly installing on Minnow Board Max. Running  Debian 8|  Not working in python|


v1v3kdalal:
After following the Basic build steps section. I try to use mraa in python. (Tested in 2.7 and 3.4) In 3.4 it does import mraa but as soon as you call a function it gives errors. In 2.7 it won't even import.
The build examples written in cpp do work however.

Any help would be appreciated, i need to use mraa in python.
I have all the python dev's installed.
Thank you. 

Here is the CMakeErrorLog

Error: http://dpaste.com/0A27YHT

arfoll:
Try use a pastebin service like dpaste intead of this github messes up all the formatting.

Our cmake config can get confused between python versions, tbh by default it should force python2 usage unless you set -DBUILDPYTHON3=ON. Wipe the build/ dir and try building again with that option and paste your error in python3.

alext-mkrs:
Yeah, the output is quite inconvenient to read. What I could derive from that is thet it can't find `pthread_create()` function.

@v1v3kdalal, please post the output of this command: `dpkg -l |egrep 'libc|pthread'`. If it's longer than a couple of lines, please use some service like dpaste.

v1v3kdalal:
@alext-mkrs We entered your command here is the output:
http://dpaste.com/37J5BCS 
@arfoll When we built the  cmake files i entered the -DBUILDPYTHON3=ON when i ran make.
However we will see what @alext-mkrs suggests then if needed we will clean the build and reinstall again making sure we use that build command when we build.

v1v3kdalal:
@arfoll We wiped our build directory. Then typed make -DBUILDPYTHON3=ON .. then make.
Everything seemed to install correctly.


Then we tried running a MRAA test code in Python 3.4 and this is the error message.
http://dpaste.com/3JM4CC0

alext-mkrs:
Ok, so the build problem is gone, now you're trying to run a python script and it sort of fails. Please post your full script, that checksMRAA.py one.

arfoll:
@v1v3kdalal I pushed a fix to python3 builds just now, actually we did have a build issue where @alext-mkrs added some useful debug info but makes a call to a function that is python2 only :/. Try update master and build again with commit 083606790061d4c8c70c66007439410ce30de063.

But considering that you are building my guess is you have another issue yet. Please paste the full output of cmake like this: http://dpaste.com/2NCJ5F0

[...]

PR #359, opened 2015-11-11
cmake: use cmake includedir

tripzero:


tripzero:
something still isn't quite right... plz wait.

arfoll:
Happy to wait but can we get more details in the commit message as to why this change is needed/wanted and what problem it solves?

tripzero:
Sure.  Will do.

tripzero:
The problem space is more complex than this patch fixes.  The issue is within yocto and cmake itself.  Within cmake, CMAKE_INSTALL_PREFIX has a unique meaning[1].  It means the same as DESTDIR when running "make DESTDIR=/foo install".  In yocto, this variable is being inappropriately set to ${prefix} which is not the same as DESTDIR (yocto probably uses this elsewhere).  The result is that many files are being installed inappropriately: /usr/etc instead of /etc, /usr/lib/systemd instead of /lib/systemd, etc.

If we fix cmake.bbclass in yocto, a bunch of packages that expect ${CMAKE_INSTALL_PREFIX} including mraa.

In mraa, we can fix this by using GnuInstallDirs or the default cmake install variables instead of hardcoding with the expectation that CMAKE_INSTALL_PREFIX is /usr.

The fact that cmake makes "/usr/local" the default prefix is disturbing to me.  I think it gives developers a false assumption about its meaning.

[1] - https://cmake.org/cmake/help/v3.2/variable/CMAKE_INSTALL_PREFIX.html

tripzero:
I changed the commit message to explain the rationale.  There's one more patch I've got coming but I will make a separate commit because I don't really like the patch and I think more discussion is required to "Do It Right" TM.

tripzero:
@arfoll I removed the nodejs change.  Any comments on the python changes?

alext-mkrs:
Actually I still see a patch with nodejs changes in this pull request, are you sure you've removed it?

tripzero:
@alext-mkrs check the commit again.  I don't see any nodejs changes.

alext-mkrs:
And now I don't see that commit either  so it's ok. I think it would also help if you describe the algorithm you want to use here.

tripzero:
The algorithm for the python site-package path finding?  My algorithm uses the getsitepackages() python call to get the OS site package path and then replaces the prefix.

The problem with the old method was that it insalled to "site-packages" on Debian (and debian-like distros) and Debian doesn't use site-packages so python can't find mraa.

src/python/CMakeLists.txt
```diff
@@ -38,12 +38,33 @@ execute_process (
...
 execute_process (
    COMMAND ${PYTHON_EXECUTABLE} -c
-       "import site, sys; sys.stdout.write(site.getusersitepackages().replace(site.getuserbase(), site.PREFIXES[-1]))"
+       "	\n
+import site, sys \n
+dirs = site.getsitepackages(); \n
+dir = dirs[0] \n
+for d in dirs: \n
+	if d.count(\"packages\"): \n
+		dir = d[d.index(\"packages\")-5:] \n
```
alext-mkrs: this "-5" will break if the directory for modules is called e.g. "usersite-packages", i.e. when the part before "packages" is not five characters long.

src/python/CMakeLists.txt
```diff
@@ -38,12 +38,33 @@ execute_process (
...
+		break \n
+py = \"${PYTHON_EXECUTABLE}\" \n
+py = py[py.rfind(\"/\")+1:] \n
+prefix = \"${PYTHON_PREFIX}\"[\"${PYTHON_PREFIX}\".rfind(\"/\"):] \n
+dir = \"{0}/lib/{1}/{2}\".format(prefix, py, dir) \n
```
alext-mkrs: `py` will not work on e.g. Edison - where the executable is `/usr/bin/python2`, but the packages dir is `/usr/lib/python2.7/site-packages`, i.e. Python executable name is not equal to `/usr/lib` subfolder.

src/python/CMakeLists.txt
```diff
@@ -38,12 +38,33 @@ execute_process (
...
 )
-file (TO_CMAKE_PATH "${PYTHON_SITE_DIR}" PYTHON_SITE_DIR)
-string (REGEX REPLACE "^${PYTHON_PREFIX}/" ""
-   PYTHON_SITE_DIR "${PYTHON_SITE_DIR}")
+#"import site, sys; sys.stdout.write(site.getsitepackages()[0])"
+
+message(STATUS "python site dirs: ${PYTHON_SITE_DIR}")
+message(STATUS "python prefix: ${PYTHON_PREFIX}")
+
+if("${CMAKE_INSTALL_PREFIX}" STREQUAL "/usr")
```
alext-mkrs: What if it's not `/usr`?

Could you please clarify the point of this whole block? The `PYTHON_SITE_DIR` variable is being set to `lib/python2.7/dist-packages` on my Debian 8 after this is executed, so the `install()` function will try to install it into current directory.

alext-mkrs:
I've just checked your patch out and while I think the correctiong in `examples/CMakeLists.txt` and `src/CMakeLists.txt` make sense, I would say the Python part might need changes, as it IMHO replaces current inoptimal approach to another one, which happens to work for current Debian, but has some drawbacks I've tried to highlight in per-line comments.

We definitely need to fix that Python install part (it was already a problem mentioned in #336, just in some more optimal way. I'll see if I can come up with some proposal as well.

tripzero:
Alext, fair enough.  I don't like the python changes and your criticisms are valid.  I will remove that part of the patch and resubmit.

arfoll:
Merged, excuse the delay!

PR #363, opened 2015-11-14
Change V8_VERSION_HEX to hexadecimal string

xbolshe:


xbolshe:
Just check V8 version: https://github.com/intel-iot-devkit/mraa/pull/297 
4.5.103.33 must be translated correcly (like hexadecimal value) and not like 0x0045103

xbolshe:
BTW: current V8 version is 4.8.294:
https://chromium.googlesource.com/v8/v8.git/+/4.8.294

It is not so easy to translate it to V8_VERSION value.

xbolshe:
See https://github.com/swig/swig/issues/561

xbolshe:
I have tried to use NodeJS V8 API and a value from a command line to detect V8_VERSION_NUMBER:

va1

va2

arfoll:
This looks really good, unfortunatly as I commented on your SWIG upstream issue there is a bug meaning this breaks 0.12.x which sadly does mean I can't really merge it unless I do a big hack ontop since there are 0.12.x users I can't annoy ehehe

arfoll:
@xbolshe With reguards to the comment from @ojwb do you still think 4.5.103.33 is not being translated correctly?

xbolshe:
Yes, I do. It may be 4.5.10.333 or 4.51.03.33 as well. My analysis of SWIG shows that all versions used in SWIG are related with VERSION_MAJOR=3. It means that moving to V8 major version equal 4 may fix several problems.

arfoll:
What I don't understand is that even with the current - poor - version conversion we would have VERSION_MAJOR=4 for 4.5.103.33. Doing it this way causes node 0.12.x support to break so I can't merge this. Closing this as I'm not clear what bug it would solve.

PR #365, opened 2015-11-17
Updates for main README file and uart.c

alext-mkrs:
First commit is based on suggestions from #336, second one is just style/formatting updates to make it nicer, third one is an Edison SDK version bump - we are at 1.7.2 now and the fourth one is probably a copy-paste artifact in `uart.c` I've noticed while investigating something recently.


arfoll:
Merged, thanks. I added UPM style API doc links at the bottom too

PR #366, opened 2015-11-21
FTDI4222 on-chip GPIO supported directly from mraa

Propanu:
So far we have the 4 on-chip GPIO's working nicely and we can now switch GPIO0 & GPIO1 between modes. Also made few other improvements that will allow new I2C expanders to be added easily.
Still left:
* SPI implementation
* Add the new GPIO expanders
* Extend ISR to support all edge modes

### DIFF
--- a/src/gpio/gpio.c
+++ b/src/gpio/gpio.c
@@ -584,6 +584,10 @@ mraa_gpio_write(mraa_gpio_context dev, int value)
             return pre_ret;
     }
 
+    if (IS_FUNC_DEFINED(dev, gpio_write_replace)) {
+        return dev->advance_func->gpio_write_replace(dev, value);
+    }
+
     if (dev->value_fp == -1) {
         if (mraa_gpio_get_valfp(dev) != MRAA_SUCCESS) {
             return MRAA_ERROR_INVALID_RESOURCE;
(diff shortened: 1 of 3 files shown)


tingleby:
So, It looks OK and it builds for me too. I don't have the chip to hand at the moments so I can't comment too much on the functionality.

One thing I have noticed is that you introduce white space in earlier patches in the PR which you then fix later, whilst the result is OK, its best that you never introduce white-space where it shouldn't be.
So if we were to only cherry-pick from a PR, the quality would still remain high. But there isn't a need to change this now, just advice for the future.

Propanu:
Updated gpio write function replacement so this should all be set now. Sorry about white-spaces, my editor auto-formats and I randomly press tab too at times :smile:.

arfoll:
set `:set expandtab` in vim ;-) Can you just rebase those 2 in the original commits?

Propanu:
There we go, on top of master too!

arfoll:
Merged. Please use more care when rebasing, it took me a good 20 minutes to clean that up so it doesn't look like a botch job. It's fine to commit quickly when doing dev but please clean it all up nicely before merging/PR into master.

PR #370, opened 2015-11-24
java: Added auto load library code to the SWIG interface file

stefan-andritoiu:


arfoll:
Merged, Thanks!

PR #373, opened 2015-11-30
Memory leak and out of bound

laykuanloon:


arfoll:
Merged from email patches- thanks

Issue #376, opened 2015-12-02
Uart python example


BTWS:
Is it possible to provide a more elaborate example for uart in python?

alext-mkrs:
Definitely! Do you have any specific scenario or use cases in mind?

BrechtW:
Communicating with a Dynamixel servo to be specific. This requires features like only writing when the input buffer isn't full. Setting baudrate, timeout, flush, read, write.

alext-mkrs:
Ok, thanks. I don't have such a servo, but the information you've provided is enough for me to get the idea.

PR #383, opened 2015-12-13
Reworked Python example for UART, updated docs

alext-mkrs:
The first commit is a full rework of the Python UART examples, per the request in #376 and the second commit adds a piece for UART into Python docs, examples section.


arfoll:
Thanks for improving that sample 100 fold :) Merged

PR #378, opened 2015-12-04
ftdi4222: minor fixes and updated documentation

Propanu:
Might want to wait on this for a bit as I will be pushing more changes over the next week.

PR #384, opened 2015-12-16
Re-ordered ft4222 sub-platform GPIO numbering

whbruce:
Built-in ft4222 gpio pin now always appear as pins 0-3.
If i2c i/o expander is detected, its pins are appended as 4-11.
Apologies in advance for using merge instead of rebase.


arfoll:
I'm not clear what I'm looking at here - there's alot of stuff, can you extract the patches one by one and stage them in a branch? Your git master is very different from this master so I'd suggest deleting it and starting again (or just using git reset). Also if this could include #378 so it's clearer what comes first it would be goood. Thanks!

src/usb/ftdi_ft4222.c
```diff
@@ -56,96 +66,87 @@ mraa_ftdi_ft4222_init()
...
-            if (FT4222_OK == ft4222Status){
-                syslog(LOG_NOTICE, "FT4222_GetVersion %08X %08X\n", ft4222Version.chipVersion,
-       ft4222Version.dllVersion);
-             } else
-                syslog(LOG_ERR, "FT4222_GetVersion failed with code %d", ft4222Status);
-    */
-
-    syslog(LOG_NOTICE, "FT_GetDeviceInfoList returned %d devices\n", numDevs);
-
+    if (numDevs < 2) {
```
Propanu: CNFMODE3 will only list 1 device. I still think we should support it and skip initializing the builtin GPIOs.

PR #385, opened 2015-12-17
uart_ow: Initial support for Dallas 1-wire over UART support for MRAA

jontrulson:
This commit introduces support for Dallas Semiconductor (DS) 1-wire
compliant device support using an available UART device.

The principle of operation is described in the following Application
note by Maxim Electronics:

https://www.maximintegrated.com/en/app-notes/index.mvp/id/214

It has been tested on Galileo 2 and Edison, with 2 DS 1-wire devices,
the DS18B20 and DS2413 connected to the bus.  A UPM driver for the
DS2413 is already complete and a PR will be submitted after this one.

Once this support has been added, a driver for the DS18B20 will be
completed (depending on priorities), and UPM support for other DS
1-wire devices can be implemented at will.

In order for this to work, a simple interface circuit, using a single
diode must be constructed:

(forgive my "Asciihematic" :)

-|
U|
A| TX---|<-+
R|         |
T| RX------+---1-wire data-----o
-|

The diode on TX is a 1N4148 (cheap and common), with the cathode
connected to TX, and the anode connected to RX and the 1-wire data
line.

The 1-wire data line requires a pull-up resistor, as the DS 1-wire
spec requires. 4.7-5K is typical for DS 1-wire buses.

NOTE: DHT-type (temp/humidity sensor) 1-wire devices ARE NOT DS
1-wire compliant, and will not work with this code/circuit.

Also note, this will use up one of your UARTs, which cannot be used
for any other purpose (ie: to access true UART-type serial devices).

You can however, connect as many DS 1-wire devices as feasible to this
UART, as it will function as a DS 1-wire bus master.


alext-mkrs:
That's a great addition. I see you state DHT-type ones won't work, is that just because of a different protocol or this UART approach has some fundamental limitations tha prevent that from happening?

jontrulson:
Yes, the issue is that the DHT's don't conform to the DS 1-wire protocol (probably because they didn't want to license it).  The DS 1-wire protocol allows the master to control the bit-time slots, plus a specific method for detecting device presence, etc.  This seems to work pretty well using a UART as the master.

The DHT's are basically a fire-and-forget.  You trigger a LOW pulse and it starts sending data with a varying time signature (50us+28us, or 50us+70us IIRC) with no control over the timing.  I tried to see if I could use that technique on them without much success (UART framing errors I suspect).

alext-mkrs:
Understood, thanks.

Propanu:
@arfoll Any news on the licenses? Can this be merged?

jontrulson:
Closing this so it can be re-implemented without the ambiguous licensing issues.

PR #389, opened 2015-12-27
JavaScript: fixes and remove useless code

xbolshe:

Using ::Copy instead ::New is more correct.

### DIFF
--- a/src/javascript/mraajs.i
+++ b/src/javascript/mraajs.i
@@ -43,11 +43,10 @@ class Spi;
 %typemap(out) uint8_t*
 {
 %#if SWIG_V8_VERSION > 0x040000
-   v8::MaybeLocal<v8::Object> objret = node::Buffer::New(v8::Isolate::GetCurrent(), (char*) result, arg3);
+   v8::MaybeLocal<v8::Object> objret = node::Buffer::Copy(v8::Isolate::GetCurrent(), (char*) result, arg3);
    free(result);
-   if(!objret.ToLocal(&$result)) {
-      SWIG_exception_fail(SWIG_ERROR, "Spi buffer failed");
-      SWIGV8_RETURN(SWIGV8_UNDEFINED());
+   if(!objret.ToLocal(&$result)){
+	SWIG_exception_fail(SWIG_ERROR, "Spi buffer failed");
    }
 %#elif SWIG_V8_VERSION > 0x032870
   $result = node::Buffer::New((char*) result, arg3);
@@ -73,7 +72,6 @@ class Spi;
    $2 = x;
    if ($2 < 0) {
        SWIG_exception_fail(SWIG_ERROR, "Positive integer expected");
-       SWIGV8_RETURN(SWIGV8_UNDEFINED());
    }
    $1 = (char*) malloc($2 * sizeof(uint8_t));
 }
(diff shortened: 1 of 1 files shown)


xbolshe:
Any code after SWIG_exception_fail will never be executed, because it will be "goto fail".

arfoll:
Very good point on the exception_fail code :) Do you have any materials I could quickly read about using ::Copy instead of ::New? What versions of node did you test with?

Thanks!

xbolshe:
Node 5.3.0. After switching to git version I have found again that SPI function response returns incorrect values while using ::New.

About materials: I used only source code of Node 5.3.0. 

Please pay your attention that ::Copy creates internal buffer with malloc,
when ::New does not. It creates a problem when a source buffer is freed.


MaybeLocal<Object> Copy(Isolate* isolate, const char* data, size_t length) {
  Environment* env = Environment::GetCurrent(isolate);
  EscapableHandleScope handle_scope(env->isolate());
  Local<Object> obj;
  if (Buffer::Copy(env, data, length).ToLocal(&obj))
    return handle_scope.Escape(obj);
  return Local<Object>();
}

MaybeLocal<Object> Copy(Environment* env, const char* data, size_t length) {
  EscapableHandleScope scope(env->isolate());

  // V8 currently only allows a maximum Typed Array index of max Smi.
  if (length > kMaxLength) {
    return Local<Object>();
  }

  void* new_data;
  if (length > 0) {
    CHECK_NE(data, nullptr);
    new_data = malloc(length);
    if (new_data == nullptr)
      return Local<Object>();
    memcpy(new_data, data, length);
  } else {
    new_data = nullptr;
  }

  Local<ArrayBuffer> ab =
    ArrayBuffer::New(env->isolate(),
        new_data,
        length,
        ArrayBufferCreationMode::kInternalized);
  Local<Uint8Array> ui = Uint8Array::New(ab, 0, length);
  Maybe<bool> mb =
      ui->SetPrototype(env->context(), env->buffer_prototype_object());
  if (mb.FromMaybe(false))
    return scope.Escape(ui);

  // Object failed to be created. Clean up resources.
  free(new_data);
  return Local<Object>();
}





MaybeLocal<Object> New(Isolate* isolate, char* data, size_t length) {
  Environment* env = Environment::GetCurrent(isolate);
  EscapableHandleScope handle_scope(env->isolate());
  Local<Object> obj;
  if (Buffer::New(env, data, length).ToLocal(&obj))
    return handle_scope.Escape(obj);
  return Local<Object>();
} 

MaybeLocal<Object> New(Environment* env, char* data, size_t length) {
  EscapableHandleScope scope(env->isolate());

  if (length > 0) {
    CHECK_NE(data, nullptr);
    CHECK(length <= kMaxLength);
  }

  Local<ArrayBuffer> ab =
      ArrayBuffer::New(env->isolate(),
                       data,
                       length,
                       ArrayBufferCreationMode::kInternalized);
  Local<Uint8Array> ui = Uint8Array::New(ab, 0, length);
  Maybe<bool> mb =
      ui->SetPrototype(env->context(), env->buffer_prototype_object());
  if (mb.FromMaybe(false))
    return scope.Escape(ui);
  return Local<Object>();
}

arfoll:
Thanks alot for taking the time to explain this to me :) Merged with ff6b304fd2225ed7147ae3e7d964fd37aa934e8f

PR #397, opened 2016-01-04
Migrate IsrCallbacks to Java's Runnable class

petreeftime:
Signed-off-by: Petre Eftime <petre.p.eftime@intel.com>

### DIFF
--- a/api/mraa/gpio.h
+++ b/api/mraa/gpio.h
@@ -44,6 +44,12 @@ extern "C" {
 #include <Python.h>
 #endif
 
+#if defined(SWIGJAVA) || defined(JAVACALLBACK)
+#include <jni.h>
+extern JavaVM *globVM;
+extern void mraa_java_isr_callback(void *);
+#endif
+
 #include <stdio.h>
 #include <pthread.h>
 
--- a/api/mraa/gpio.hpp
+++ b/api/mraa/gpio.hpp
@@ -69,30 +69,6 @@ typedef enum {
     EDGE_FALLING = 3 /**< Interupt on falling only */
 } Edge;
 
-#if defined(SWIGJAVA)
-
-class IsrCallback
-{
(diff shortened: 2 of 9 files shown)


alext-mkrs:
Looks like you have a commit message comprised of several ones with multiple "signed-off-by" lines and actual commit messages glued into one, which looks a bit strange. I'd suggest you to correct that, either by sending several commits as-is, or editing the commit message so that it looks as the one for a single commit.

petreeftime:
Sure. I think those were generated from squashing multiple commits into a single one.

arfoll:
This seems to break node.js ISRs (at least under node 4.x). I'm guessing this commit https://github.com/petreeftime/mraa/commit/9cfd54268e55ca427de664078d93896b44fa9d5f#diff-9763aeef44b83561dc5bc04f11e161ebR49 is at fault

petreeftime:
That function was previously not visible at all when running SWIG, no sure why having it in the header would break ISRs on anything, but I'll look into it. What exactly is the issue and how can I replicate it?

arfoll:
Maybe my quick test was incorrect, just try run an isr from an Edison board
using node. I can test later but currently away from my boards

petreeftime:
I tested ISR on node and python and they work fine for me. The isr.js example is broken for me though as it exits immediately. On Galileo Gen2, with pin 6 instead of 14 and added this code at the end to prevent it from exiting immediately isr.js works fine, and so does isr.py:
```javascript
function periodicActivity()
{ 
  setTimeout(periodicActivity,1000); //call the indicated function after 1 second (1000 milliseconds)
}

periodicActivity(); //call the periodicActivity function
```

arfoll:
Thanks for reassuring me, apologies for the doubts!

Thanks for fixing the msg - this is merged!

PR #399, opened 2016-01-05
do not require a C++ compiler

pieterjanc:
By default, CMake assumes both a C and C++ compiler are required.
This is not needed for mraa because it only provides C++ header
files.

Signed-off-by: Pieterjan Camerlynck <pieterjan.camerlynck@gmail.com>

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,5 +1,5 @@
 cmake_minimum_required (VERSION 2.8)
-project (mraa)
+project (mraa C)
 
 FIND_PACKAGE (Threads REQUIRED)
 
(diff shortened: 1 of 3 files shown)


tingleby:
Thanks, Yeah sure why not!
Im concerned that it disables CXX even when its needed for the Swig compiles aspects of mraa. See the failed Travis build.
Conditionally adding CXX language feature to the project would probably satisfy this.
https://cmake.org/cmake/help/v3.0/command/enable_language.html?highlight=languages

pieterjanc:
Oops, built without the examples. Should be fixed now.

tingleby:
Great that looks like it might be it! It would be best as a single commit really.
I will give it a bit of testing in the morning!

pieterjanc:
It seems that @arfoll committed the patch (https://github.com/intel-iot-devkit/mraa/commit/a87a104efeb83d4e762992e5be3b8908e6d2bf40), so I'm closing the pull request.

arfoll:
Sorry I forgot to get back to you once pushed. Thanks for closing & the patch!

PR #400, opened 2016-01-05
Ft4222 gpio

whbruce:
Re-ordered ft4222 gpio so that built-in gpio id starts at 0 and i2c expander gpio comes after. 
Added support for i2c switch by exposing downstream busses as extra mraa i2c entries.
Also added some uart fixes.


arfoll:
Merged, thanks!

PR #405, opened 2016-01-17
Using calloc over malloc where appropriate

anonymouse64:
Switch to using calloc on all calls to malloc where the memory isn't initialized. For things like the mraa_board_t, not allocating all to zero causes issues such as with the sub_platform member, where if that's not zero mraa_get_platform_type will try to dereference a random memory location for the sub_platform->platform_name, which can result in segmentation faults and other issues.  

Note that in some places where immediately after the malloc call is a copy operation, there is no need for calloc, as all the memory gets overwritten anyways, but in cases where there may or may not be memory written to (such as in [mraa_file_contains](https://github.com/intel-iot-devkit/mraa/blob/master/src/mraa.c#L581), with reading from a file), even though in most cases the memory is overwritten, it could be the case that the read operation does nothing, but the memory still has non-zero values, by virtue of the fact it wasn't overwritten.

In files : 
* mraa.c
* aio.c
* 96boards.c
* banana.c
* beaglebone.c
* raspberry_pi.c
* glob.c
* i2c.c
* pwm.c
* ft4222.c
* intel_de3815.c
* intel_edison_fab_c.c
* intel_galileo_rev_d.c
* intel_galileo_rev_g.c
* intel_minnow_byt_compatible.c
* intel_nuc5.c
* intel_sofia_3gr.c

### DIFF
--- a/src/aio/aio.c
+++ b/src/aio/aio.c
@@ -58,7 +58,7 @@ aio_get_valid_fp(mraa_aio_context dev)
 static mraa_aio_context
 mraa_aio_init_internal(mraa_adv_func_t* func_table)
 {
-    mraa_aio_context dev = malloc(sizeof(struct _aio));
+    mraa_aio_context dev = calloc(1,sizeof(struct _aio));
     if (dev == NULL) {
         return NULL;
     }
--- a/src/arm/96boards.c
+++ b/src/arm/96boards.c
@@ -57,7 +57,7 @@ mraa_board_t* mraa_96boards()
    char ch;
    char name[MRAA_PIN_NAME_SIZE];
 
-   mraa_board_t* b = (mraa_board_t*) malloc(sizeof(mraa_board_t));
+   mraa_board_t* b = (mraa_board_t*) calloc(1,sizeof(mraa_board_t));
    if (b == NULL) {
       return NULL;
    }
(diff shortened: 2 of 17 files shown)


arfoll:
Merged with ab14b9de43d41774b55242099d7c0374017507a1

I changed your commit msg to match your very in depth explanantion here, since that way it'll never get lost. I also changed the prefix to something more reasonable :)

One change I did make is I didn't keep the change to glob.c, I wasn't entirely sure (newsize*char*) seems like alot more memory than before so I didn't really understand and touching memory allocation routines in such a widely used piece of code didn't seem like the smart thing to do without more testing on my side ot reassure myself that I wasn't breaking it. I'm happy to merge it if you can show me the size allocated is actually the same (it could be I didn't look all that hard at the code truth be told).

anonymouse64:
To be perfectly honest I struggled with looking at that [``glob.c``](https://github.com/intel-iot-devkit/mraa/blob/master/src/glob/glob.c#L902) code myself for quite a while trying to figure out if it was necessary. I'm still hazy on how all the glob code works, so I'm fine with leaving it as is if it's not causing any problems.

arfoll:
I'm glad I'm not the only one that gets confused at that one, I'll close this then :) Thanks again for the patch!

Issue #408, opened 2016-01-19
gcc compile error when building Node.JS binding on Fedora 22


whbruce:
I'm getting a gcc error when building Node.JS binding on Fedora 22
cmake completes successfully and relevant output is as follows:
```
INFO - libmraa Version v0.9.0-18-g7e30224
INFO - Target arch is x86_64
-- The CXX compiler identification is GNU 5.1.1
-- Check for working CXX compiler: /usr/lib64/ccache/c++
-- Check for working CXX compiler: /usr/lib64/ccache/c++ -- works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Found SWIG: /usr/lib64/ccache/swig (found version "3.0.7") 
-- Found Node: /usr/bin/node  
INFO - Node version is 0.10.36
INFO - Node using v8 3.14.5.10
DEBUG - Padded V8 version to match SWIG format
```

But when I build I see the following error
```
cd /home/hbruce/Projects/mraa/build/fc-js/src/javascript && /usr/lib64/ccache/c++   -Dmraajs_EXPORTS -g -fPIC -I/home
/hbruce/Projects/mraa/api -I/home/hbruce/Projects/mraa/api/mraa -I/home/hbruce/Projects/mraa/include -I/usr/include/s
rc -I/usr/include/node -I/usr/include/deps/v8/include -I/usr/include/deps/uv/include -I/home/hbruce/Projects/mraa/src
/javascript/..     -DBUILDING_NODE_EXTENSION -DSWIGJAVASCRIPT=TRUE -o CMakeFiles/mraajs.dir/mraajsJAVASCRIPT_wrap.cxx
.o -c /home/hbruce/Projects/mraa/build/fc-js/src/javascript/mraajsJAVASCRIPT_wrap.cxx

In file included from /home/hbruce/Projects/mraa/build/fc-js/src/javascript/mraajsJAVASCRIPT_wrap.cxx:169:0:
/usr/include/node/node.h:150:46: error: ‘uv_statbuf_t’ does not name a type
 v8::Local<v8::Object> BuildStatsObject(const uv_statbuf_t* s);
                                              ^
In file included from /home/hbruce/Projects/mraa/build/fc-js/src/javascript/mraajsJAVASCRIPT_wrap.cxx:169:0:
/usr/include/node/node.h:252:16: error: variable or field ‘SetErrno’ declared void
[...]

PR #409, opened 2016-01-19
nodejs: Detect compat-libuv010 package and use it instead of libuv

whbruce:
Fixed Fedora22 Javascript build problem as per issue #408


whbruce:
Cancel this. I hadn't noticed that you have you own FindNodeJs module. It would be better to change logic in here rather than in Javascript Cmake file. This way we can share fix with UPM project.

PR #412, opened 2016-01-22
pwm: Check for out of range pin number in pwm_init()

whbruce:
pwm_init() does not check for out of range pin number so can  cause a seg fault.


arfoll:
Thanks for the great fix!

PR #413, opened 2016-01-22
java: Ignored function, which should not have bindings created for it

stefan-andritoiu:


arfoll:
Thanks for fixing this so fast :)

Issue #410, opened 2016-01-21
cmake FindNodejs module should indicate that required include files have not been detected 


whbruce:
When the cmake option `BUILDSWIGNODE` is set, the `FindNodejs` cmake module is used to detect the presence of packages requires to build the Javascript binding. The current logic looks for the include files but no action is taken if they are not found. Thus the g++ include path is invalid and the build fails (can't find `node.h`). I hit this problem as package `nodejs` was installed but `nodejs-devel` was not.

I can update the `FindNodejs` cmake module to catch this condition and flag an error but wanted to check if current behavior was required before proceeding. 

Let me know.

alext-mkrs:
This was implemented by PR #414 for a while now - closing this one.

PR #414, opened 2016-01-22
Updated FindNodejs cmake module

whbruce:
This pull request address issues #408 and #410. 
If BUILDSWIGNODE is set and Nodejs development files are not detected, cmake will now stop processing. I have also added support for Fedora 22 via compat-libuv010 package.
I have tested on the following distros:
* Ubuntu 14.04
* Fedora 21
* Fedora 22

### DIFF
--- a/cmake/modules/FindNodejs.cmake
+++ b/cmake/modules/FindNodejs.cmake
@@ -1,14 +1,12 @@
- # Macro to add directory to NODE_INCLUDE_DIRS if it exists and is not /usr/include
+ # Macro to add directory to NODEJS_INCLUDE_DIRS if it exists and is not /usr/include
  macro(add_include_dir dir)
     if (IS_DIRECTORY ${dir} AND NOT ${dir} STREQUAL "/usr/include")
-      set(NODE_INCLUDE_DIRS ${NODE_INCLUDE_DIRS} ${dir})
+      set(NODEJS_INCLUDE_DIRS ${NODEJS_INCLUDE_DIRS} ${dir})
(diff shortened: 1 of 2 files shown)


whbruce:
@arfoll I just ran a quick test and if REQUIRED is not set when calling find_package (Nodejs), cmake will display error, but continue processing so resulting makefile will not build.

```
-- Found SWIG: /usr/lib64/ccache/swig (found version "3.0.7") 
ERROR - node.h not found
-- Could NOT find Nodejs (missing:  NODEJS_INCLUDE_DIRS) 
INFO - Node version is 0.10.36
INFO - Node using v8 3.14.5.10
DEBUG - Padded V8 version to match SWIG format
-- Configuring done
-- Generating done
```

Your call on whether REQUIRED is used or not.

arfoll:
What I meant is I like the REQUIRED bit, but the if statement behind it seems unnecessary. since having REQUIRED will still tell you:
```
-- Could NOT find Node (missing:  NODE_EXECUTABLE) 
```

which is more informative and tells you everything you know than the generic:
```
ERROR - Nodejs not found
```

whbruce:
The message "ERROR - node.h not found" is intended to show the reason why Nodejs detection failed. There are similar errors for uv.h and v8.h. I can remove these if they are providing much output.

whbruce:
"Node executable not found" message was indeed redundant. This has been removed and commit pushed.

arfoll:
Woops, this has been merged for a while - closing :)

PR #415, opened 2016-01-22
uart_ow: Initial support for Dallas 1-wire over UART support for MRAA

jontrulson:
This commit introduces support for Dallas Semiconductor (DS) 1-wire
compliant device support using an available UART device.

The principle of operation is described in the following Application
note by Maxim Electronics:

https://www.maximintegrated.com/en/app-notes/index.mvp/id/214

It has been tested on Galileo 2 and Edison, with 2 DS 1-wire devices,
the DS18B20 and DS2413 connected to the bus.  A UPM driver for the
DS2413 is already complete and a PR will be submitted after this one.

It is important that you use a UART with CMOS/TTL level voltages
(3.3v/5v) RX and TX lines.  DO NOT use standard RS232 level voltages
or you are going to have a bad day.

In order for this to work, a simple interface circuit, using a single
diode must be constructed:

(forgive my "Asciihematic" :)

-|
U|
A| TX---|<--+
R|          |
T| RX-------o--------o 1-wire data bus
-|

The diode on TX is a 1N4148 (cheap and common), with the cathode
connected to TX, and the anode connected to RX and the 1-wire data
line.

The 1-wire data line requires a pull-up resistor, as the DS 1-wire
spec requires. 4.7-5K is typical for DS 1-wire buses.

NOTE: DHT-type (temp/humidity sensor) 1-wire devices ARE NOT DS 1-wire
compliant, and will not work with this code/circuit unfortunately.

Also note, this will use up one of your UARTs, which cannot be used
for any other purpose (ie: to access true UART-type serial devices).

You can however, connect as many DS 1-wire devices as feasible to this
UART, as it will function as a DS 1-wire bus master.


Propanu:
@arfoll my understanding was that this was updated and should be now ready to go? Anything else that needs to be done here?

arfoll:
This is merged, I must have failed to close it :/

PR #422, opened 2016-02-05
Spi: Constructor with chip select added in cpp header

smuellener:
Signed-off-by: Steven Müllener <steven.muellener@sius.com>


arfoll:
Merged, thanks alot!

PR #423, opened 2016-02-06
intel_edison_fab_c.c: Enhanced detection of Arduino expansion board

alext-mkrs:
I've looked into checking for ADC as well, but on the miniboard the driver is also active and to my surprise everywhere I checked it looked like the ADC is there.

The only exception was that on the miniboard reads from `in_*_raw` files were returning zeroes and on the Arduino board [floating] pins were returning various random values. But that's too weak of a reference - if my setup had analog pins tied to GND it would return zeroes on Arduino as well.

So I've ended up with using GPIO expanders as an additional reference and I've refactored the detection out into a dedicated function to contain the logic.

### DIFF
--- a/src/x86/intel_edison_fab_c.c
+++ b/src/x86/intel_edison_fab_c.c
@@ -29,6 +29,7 @@
 #include <sys/mman.h>
 #include <sys/ioctl.h>
 #include <linux/spi/spidev.h>
+#include <errno.h>
 
 #include "common.h"
 #include "x86/intel_edison_fab_c.h"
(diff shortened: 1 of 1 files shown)


alext-mkrs:
That's how it looks like in the log.

Arduino board:
```
Feb 06 23:11:22 edison1 libmraa[883]: libmraa version v0.9.1-5-g3e1c528 initialised by user 'root' with EUID 0
Feb 06 23:11:22 edison1 libmraa[883]: edison: Arduino board detected
Feb 06 23:11:22 edison1 libmraa[883]: libmraa initialised for platform 'Intel Edison' of type 2
```

Miniboard:
```
Feb 06 23:44:25 edison2 libmraa[909]: libmraa version v0.9.1-6-g44b5c21 initialised by user 'root' with EUID 0
Feb 06 23:44:25 edison2 libmraa[909]: gpio: Failed to write 214 to export
Feb 06 23:44:25 edison2 libmraa[909]: edison: tristate not detected
Feb 06 23:44:25 edison2 libmraa[909]: edison: Arduino board not detected, assuming Intel Edison Miniboard
Feb 06 23:44:25 edison2 libmraa[909]: libmraa initialised for platform 'Intel Edison' of type 2
```

alext-mkrs:
Just realized `clang-format` run might come in handy and looks like it would change some of the formatting in my patch, but also some other text in `intel_edison_fab_c.c` (beyond my changes). Let me know if you want me to reformat just my piece or everything (and split that into two commits).

arfoll:
Thanks for this, I want to test it a little more and will merge. Styling fixes aren't necessary, they complicate the history for no good reason. I need to be more diligent when merging patches but that's on me :/

alext-mkrs:
Sure, thanks

Propanu:
@alext-mkrs I was under the impression that the ADS7955 driver doesn't get loaded on the Minibreakout since it doesn't detect the chip. Is that not true? This might be a bit off topic but the implication here would be that SPI will be just as bad on the Minibreakout because of that.

alext-mkrs:
It looks like the driver is not loaded as a module, but is statically built into kernel (so it's always there) and it does activate all the relevant endpoints under /sys/class/iio, same as on the Arduino board, despite hardware not being available. I've thoroughly checked both and observed that.

arfoll:
This makes sense, as much a some people would like to pretend, you can't autodetect devices on a SPI bus so the driver gets loaded and then apart from locking a chip select it's essentially useless.

arfoll:
Merged, thanks!

PR #424, opened 2016-02-07
js examples: removed surplus char() helper function

alext-mkrs:
@arfoll, if you don't mind as the owner of that bug record :smiley: 

I've tested bmp85 and spi examples (not sure what should I have for rgblcd, Grove LCD?) to produce the same results on node 0.10.38 (default in the image) and 0.12.7 (from my repo).

### DIFF
--- a/examples/javascript/bmp85.js
+++ b/examples/javascript/bmp85.js
@@ -26,30 +26,27 @@
 
 var m = require('mraa'); //require mraa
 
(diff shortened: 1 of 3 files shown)


arfoll:
Merged - cheers. c101117ecedf97a2f1ed0156de32dec547ab6138

Tbh I still can't remember why those where needed... I swear it didn't work without!

PR #427, opened 2016-02-12
ft4222: added full support for PCA9555 read/write

Propanu:
With memory mapped writes. Detection could use some rework but as is will work in 99% of the cases.


arfoll:
is merged

PR #428, opened 2016-02-13
intel_edison_fab_c.c: Fixed file descriptor leak in case of read error

alext-mkrs:
There's a small bug in my previous Arduino board detection patch - in case of GPIO expander label read error we aren't closing the label file properly, which should be fixed.


alext-mkrs:
Merged in baa1fcf7be699662077a59a899f1e51ac8a6bb9f, closing the PR.

Issue #433, opened 2016-02-17
Unable to use both UPM & MRAA with FT4222


Propanu:
Title says it all, if you try to use both MRAA and some UPM driver with the FT4222, the second subplatform init will fail. So far only tested this with node but I assume it will happen in other languages too.
Here's the code:
```javascript
> var mraa = require('mraa');
undefined
> var upm = require('jsupm_stepmotor');
undefined
> var sm = new upm.StepMotor(516, 517);
Error: Illegal arguments for construction of _exports_StepMotor
    at repl:1:10
    at REPLServer.self.eval (repl.js:110:21)
    at repl.js:249:20
    at REPLServer.self.eval (repl.js:122:7)
    at Interface.<anonymous> (repl.js:239:12)
    at Interface.emit (events.js:95:17)
    at Interface._onLine (readline.js:203:10)
    at Interface._line (readline.js:532:8)
    at Interface._ttyWrite (readline.js:761:14)
    at ReadStream.onkeypress (readline.js:100:10)
> var pin = new mraa.Gpio(516);
undefined

```
And here's the log:
```sh
Feb 18 07:34:44 WR-IDP-F6DA libmraa[4941]: libmraa version v0.9.3-1-g64c162c initialised by user 'root' with EUID 0
Feb 18 07:34:44 WR-IDP-F6DA libmraa[4941]: Platform not supported, not initialising
Feb 18 07:34:44 WR-IDP-F6DA libmraa[4941]: FT_GetDeviceInfoList returned 2 devices
Feb 18 07:34:44 WR-IDP-F6DA libmraa[4941]: mraa_ftdi_ft4222_init completed successfully
Feb 18 07:34:44 WR-IDP-F6DA libmraa[4941]: FT4222_GetVersion 42220100 01020104
Feb 18 07:34:44 WR-IDP-F6DA libmraa[4941]: ft4222: using pca9555 as GPIO expander
Feb 18 07:34:44 WR-IDP-F6DA libmraa[4941]: libmraa initialised for platform 'Unknown platform + FTDI FT4222' of type 98
Feb 18 07:34:54 WR-IDP-F6DA libmraa[4941]: libmraa version v0.9.3-1-g64c162c initialised by user 'root' with EUID 0
Feb 18 07:34:54 WR-IDP-F6DA libmraa[4941]: Platform not supported, not initialising
[...]

PR #441, opened 2016-02-25
JavaScript, Java and Python bindings now use libmraa.so instread of statically linking

whbruce:
This PR address #433 for JavaScript and Java, but no Python.  
0041801 is the Java update and is a significant change; look at commit message for details.
@petreeftime, please get Java team to review this update.

### DIFF
--- a/api/mraa/gpio.h
+++ b/api/mraa/gpio.h
@@ -44,15 +44,8 @@ extern "C" {
 #include <Python.h>
 #endif
 
-#if defined(SWIGJAVA) || defined(JAVACALLBACK)
-#include <jni.h>
-extern JavaVM *globVM;
-extern void mraa_java_isr_callback(void *);
-#endif
-
 #include <stdio.h>
 #include <pthread.h>
-
 #include "common.h"
 
 /**
--- a/api/mraa/gpio.h
+++ b/api/mraa/gpio.h
@@ -40,10 +40,6 @@
 extern "C" {
 #endif
 
-#ifdef SWIGPYTHON
-#include <Python.h>
-#endif
-
 #include <stdio.h>
 #include <pthread.h>
 #include "common.h"
--- a/api/mraa/gpio.hpp
+++ b/api/mraa/gpio.hpp
@@ -28,6 +28,13 @@
 #include "types.hpp"
 #include <stdexcept>
 
+#if defined(SWIGJAVA) || defined(JAVACALLBACK)
+extern "C" {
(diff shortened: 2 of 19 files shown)


arfoll:
Why is python not done? Any reason?

whbruce:
I've never written any python code before so wouldn't feel comfortable making the update.  I guess I've got to get to grips with python sometime and it might as well be now. Just don't expect a quick fix....

arfoll:
It should be really easy just requires the same thing as for nodejs. DW if
it doesn't work I'm happy to add it was just wondering if there was a
technical reason

whbruce:
OK. Will give it a go.

whbruce:
I got the Python binding working with dynamic linking but I realized a neater solution would be to use to adv_function table as an entry point for all binding specific code for the GPIO ISR. This would remove the binding language #ifdefs in gpio.c and make the code much neater. I'l work on this tomorrow for all bindings and hope to update PR by early next week.

arfoll:
this likely won't work in node.js since that requires it to be done from C++ at least in the current methodology.

whbruce:
I've left c++ code as is. It was the lack of Javascript native code in gpio.c that made the binding update so easy. No so luck with Java and Python. I'm testing with GPIO ISR samples for all languages to ensure my changes don't break anything.

whbruce:
I have pushed updates for Java and added Python support. I enhanced gpio ISR samples so I could test with a specified pin and exercise shutdown.

petreeftime:
The patch works for Java. I checked with the FTDITest example, but made use of UPM for the button and led.

arfoll:
This is merged into the javacheck2 branch, please use that as the base for future commits. I still see some issues:

1. gpio raw mode is broken if we don't have a platform, setting lang_func needs to not depend on mraa_init() returning MRAA_SUCCESS or MRAA_ERROR_PLATFORM_ALREADY_INITIALISED for example MRAA_ERROR_PLATFORM_NOT_INITIALISED is perfectly valid. mraa_init() likely needs a small rework to return MRAA_SUCCESS to always create a MRAA_NULL_PLATFORM instead of just in USBPLAT usecase a little like done in the firmata branch.
2. This needs to be coverity clean, I added a check to the calloc on mraa_init for lang_func but I'm guessing it's not the only one... can someone run coverity or at least clang-analyzer and clean up the errors? Ping me if you need a hand there to run it in offline mode.

arfoll:
This is basically all merged now. Thanks!

Issue #403, opened 2016-01-08
Segmentation fault while using interrupt


sonofaforester:
My program receives an interrupt and then reads from the spi bus. While running, usually after 30 seconds or so, I'll get a segmentation fault. I'm using mraa v0.9.0 and node v0.12.7.

Using segfault-handler, I get the following:

PID 719 received SIGSEGV for address: 0x6960001c
/kcftech/socket/bin/node_modules/segfault-handler/build/Release/segfault-handler.node(+0x1288)[0xb77df288]
linux-gate.so.1(__kernel_rt_sigreturn+0x0)[0xb77e240c]
node(_ZN2v88Function4CallENS_6HandleINS_5ValueEEEiPS3_+0x13)[0x8191ab3]
/usr/lib/node_modules/mraa/mraa.node(_ZN4mraa4Gpio5v8isrEP9uv_work_si+0x118)[0xb6f97f21]
node(uv__work_done+0x86)[0x8557cb6]
node[0x85597d6]
node[0x8559885]
node(uv__io_poll+0x393)[0x8566273]
node(uv_run+0x112)[0x855a132]
node(_ZN4node5StartEiPPc+0x18f)[0x84ee3bf]
node(main+0x1b)[0x817295b]
/lib/libc.so.6(__libc_start_main+0xd9)[0x4e585b59]
node[0x8172c1c]

arfoll:
Do you get the segfault when the interrupt is fired or randomly? How often does the interrupt fire? Just trying to get data so I can reproduce.

Also you may want to quickly try using master, ff6b304fd2225ed7147ae3e7d964fd37aa934e8f may fix some memory allocations on 0.12.x node.js.

alext-mkrs:
Also, what do you see in the system log, `journalctl -xn 50 --no-pager` after it crashes?

sonofaforester:
Thanks for looking at this. Let me get back to you, I don't have the hardware to test it on this week.

To answer the earlier question, pulled master branch on 1/7/2016 and had the same problem. The interrupt fires about every 100ms, can't say if the segfault is on the interrupt or random but it only happens when receiving interrupts. When I get my hardware back, I'll test.

alext-mkrs:
Thanks. BTW if you're able to post your code (or better yet a reproducer - as short as possible) along with your HW description, that should help troubleshooting a lot.

alext-mkrs:
@sonofaforester so how about logs and ideally a reproducer for this one?

[...]

PR #446, opened 2016-02-28
gpio.hpp: added HandleScope creation for Node 4+

### LINKED ISSUE #387: FATAL ERROR: v8::HandleScope::CreateHandle() with nodejs (4.2.3) and button.isr()
Hey there,
I'm totally new to the hole microprocessor stuff so there might be a good chance that I'm missing something.

I've installed mraa 0.9.0 on my edison and I'm using it with the [sparkfun OLED block] (https://learn.sparkfun.com/tutorials/sparkfun-blocks-for-intel-edison---oled-block-?_ga=1.205389012.595082432.1450255418).

Currently I'm trying to read the button state of that block by following the [isr.js](https://github.com/intel-iot-devkit/mraa/blob/master/examples/javascript/isr.js) example.

When I run that program and press the corresponding button, the program crashes with:
`FATAL ERROR: v8::HandleScope::CreateHandle() Cannot create a handle without a HandleScope`

What am I missing? The code is really simple and should work:

```javascript
var m = require('mraa');

function h(args) {
  console.log("HELLO!!!!", args);
}

x = new m.Gpio(m.INTEL_EDISON_GP49);
x.isr(m.EDGE_BOTH, h);

(function loop() {
  setTimeout(loop, 1000);
})();
```

Thanks in advance.

alext-mkrs:
Closes #387.

### DIFF
--- a/api/mraa/gpio.hpp
+++ b/api/mraa/gpio.hpp
@@ -136,6 +136,9 @@ class Gpio
     static void
     v8isr(uv_work_t* req, int status)
     {
+#if NODE_MODULE_VERSION >= 0x000D
+        v8::HandleScope scope(v8::Isolate::GetCurrent());
+#endif
         mraa::Gpio* This = (mraa::Gpio*) req->data;
         int argc = 1;
         v8::Local<v8::Value> argv[] = { SWIGV8_INTEGER_NEW(-1) };


alext-mkrs:
Just in case - the proposed `Nan::EscapableHandleScope` requires additional `nan` module and that looked like an overkill to me. Also, the real fault was not in the `isr()` function, but `v8isr()` one.

Interestingly, adding this makes segmentation fault observed in #403 appear later - ~10 minutes vs ~3 minutes on Node.js 0.12.7.

arfoll:
Merged thanks alot for looking into this - very interesting & weird.

PR #458, opened 2016-03-11
WIP: PWM for Firmata

Propanu:
Still need to test and wrap up functions, but at this point suggestion/comments/concerns are welcome.

### DIFF
--- a/src/firmata/firmata_mraa.c
+++ b/src/firmata/firmata_mraa.c
@@ -414,6 +414,38 @@ mraa_firmata_gpio_dir_replace(mraa_gpio_context dev, mraa_gpio_dir_t dir)
     return MRAA_SUCCESS;
 }
 
+static mraa_pwm_context
+mraa_firmata_pwm_init_internal_replace(void* func_table, int pin)
+{
+    mraa_pwm_context dev = (mraa_pwm_context) calloc(1, sizeof(struct _pwm));
+    if (dev == NULL) {
+        return NULL;
+    }
+    dev->pin = pin;
+    dev->period = 2048000; // Locked, in ns
+    dev->advance_func = (mraa_adv_func_t*) func_table;
+
+    return dev;
+}
+
+static mraa_result_t
+mraa_firmata_pwm_write_replace(mraa_pwm_context dev, float percentage)
+{
+    return MRAA_ERROR_FEATURE_NOT_IMPLEMENTED;
+}
+
+static float
+mraa_firmata_pwm_read_replace(mraa_pwm_context dev)
+{
+    return MRAA_ERROR_FEATURE_NOT_IMPLEMENTED;
+}
+
+static mraa_result_t
+mraa_firmata_pwm_enable_replace(mraa_pwm_context dev, int enable)
+{
+    return MRAA_ERROR_FEATURE_NOT_IMPLEMENTED;
+}
+
 static void*
 mraa_firmata_pull_handler(void* vp)
 {
@@ -451,6 +483,8 @@ mraa_firmata_plat_init(const char* uart_dev)
     b->i2c_bus_count = 1;
     b->def_i2c_bus = 0;
     b->i2c_bus[0].bus_id = 0;
+    b->pwm_min_period = 2048000;
+    b->pwm_max_period = 2048000;
 
     b->pins = (mraa_pininfo_t*) calloc(b->phy_pin_count, sizeof(mraa_pininfo_t));
     if (b->pins == NULL) {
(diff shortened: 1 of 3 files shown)


arfoll:
Merged. thx

src/firmata/firmata_mraa.c
```diff
@@ -451,6 +483,8 @@ mraa_firmata_plat_init(const char* uart_dev)
     b->i2c_bus_count = 1;
     b->def_i2c_bus = 0;
     b->i2c_bus[0].bus_id = 0;
+    b->pwm_min_period = 2048000;
+    b->pwm_max_period = 2048000;
```
rwaldron: Is this a typo?

Propanu: No, it's because you cannot change the pwm period through Firmata.

rwaldron: Yes I'm very familiar with the Firmata protocol, but it's unclear that this is related to any constraint of the protocol. 

I asked if it was a typo because this is an incredibly large period in µS to used as a "minimum" (it's over 2 seconds long!). What is the basis/source of this particular value? Based on all other [occurrences](https://github.com/intel-iot-devkit/mraa/search?utf8=%E2%9C%93&q=pwm_max_period) of these properties, a smaller min/max or default would be appropriate.

arfoll:
@rwaldron thanks for having a look - so is it possible to call setPwmFrequency() or similar via Firmata? Basically the reason it's set like that is simply so it can't be changed by users and to give a sensible value when someone asks for the limits and since it can't be changed setting it this way works. That period is the 101 default period - https://github.com/01org/corelibs-arduino101/blob/master/variants/arduino_101/variant.h#L94

To be honest maybe a better move would be to respond MRAA_ERROR_FEATURE_NOT_SUPPORTED when such functions are called using the Firmata backend. What do you think?

rwaldron:
> That period is the 101 default period - https://github.com/01org/corelibs-arduino101/blob/master/variants/arduino_101/variant.h#L94

Yes, but that's a value in _nanoseconds_, not _microseconds_. Looking at that line and the line above it: 

```
#define PWM_FREQUENCY		490
#define PWM_PERIOD_NS       2048000
```

1000ms / 490 = 
2.0408ms = 
2040.8µs = 
2040800ns

But the value in this implementation is used in terms of µs: https://github.com/intel-iot-devkit/mraa/blob/firmata/src/pwm/pwm.c#L341-L349

```c
mraa_result_t
mraa_pwm_period_us(mraa_pwm_context dev, int us)
{
    if (us < plat->pwm_min_period || us > plat->pwm_max_period) {
        syslog(LOG_ERR, "pwm: period value outside platform range");
        return MRAA_ERROR_INVALID_PARAMETER;
    }
    return mraa_pwm_write_period(dev, us * 1000);
}
```

The [mraa internal type `mraa_board_t` struct defines has comments indicating that the value is µs](https://github.com/intel-iot-devkit/mraa/blob/8b68f3040e8a3e9706e52636f194082dd47d75aa/include/mraa_internal_types.h#L283-L285)

```c
int pwm_default_period; /**< The default PWM period is US */
int pwm_max_period; /**< Maximum period in us */
int pwm_min_period; /**< Minimum period in us */
```

(It looks like some platforms might be using ns for their `pwm_max_period`: https://github.com/intel-iot-devkit/mraa/search?p=2&q=period&utf8=%E2%9C%93)

rwaldron:
>  thanks for having a look - so is it possible to call setPwmFrequency() or similar via Firmata? 

No, the defaults are used: 

- Servos: [#define REFRESH_INTERVAL 20000// minumim time to refresh servos in microseconds](https://github.com/arduino/Arduino/blob/435fc323e0ab3b03bc5835e5309b5022bf57388a/libraries/Servo/src/Servo.h#L77)
- "analogWrite" (whatever the default is for the pin & timer being used)

Propanu:
I was under the impression that MRAA works with a period in nanoseconds too, as indicated by `mraa_pwm_write_period(dev, us * 1000)` in `mraa_pwm_period_us`. Perhaps the comments in mraa internal types are off, @arfoll can you confirm this?

Also, the return code was added temporarily. I will provide a basic implementation for those functions that will at least allow you to change the duty cycle using Firmata's analogWrite.

arfoll:
@Propanu no the comments from @rwaldron are sport on, the value is wrong in your code.

Propanu:
Yeah, my bad, and it's obvious from `us < plat->pwm_min_period || us > plat->pwm_max_period` which I missed. They are microseconds, like mentioned in the comments. However `dev->period` is in nanoseconds hence why this can get confusing.

PR #459, opened 2016-03-12
Change pin initialization and fix Galileo Gen2 AIO

xbolshe:
1. add pin commands. Old mraa_setup_mux_mapped style is used for other boards (pincmd = PINCMD_UNDEFINED) where pincmds are not defined.
2. remove useless pullup_enable code
3. Galileo Gen2: review all pin mux and add commands
4. Galileo Gen2: remove doubled functionality: mraa_intel_galileo_gen2_i2c_init_pre and mraa_intel_galileo_gen2_uart_init_pre
5. Galileo Gen2: fix "Invalid AIO pin specified - do you have an ADC?" error
6. Galileo Gen2: pullup/pulldown resistors are disabled during UART/GPIO/SPI/I2C/UIO initialization. Use mraa_gpio_mode to enable resistors.


arfoll:
I've temporarily queued this into qatest branch to make sure functionally nothing is broken, apart from the simple changes mentioned I think this is good

xbolshe:
"Can we make this 0xFF, there's no need to make an enum that big"

Actually it does not matter:

mraa_c

xbolshe:
" but the return value of check_ret_code is never checked for example it does not seem necessary"

I guess that it is nessesary, because if there is an error during pin's direction/value/mode setup in a Linux driver, an application will be stopped with an error. It is good for debugging of MRAA.

BR,
xbolshe

arfoll:
Thanks for putting the work in to fix it all up. I merged your 3rd commit into the first one and merged

PR #465, opened 2016-03-16
uart: Fix for failed malloc

pylbert:
Check malloc return in UART read.  Throw if malloc
fails.


arfoll:
Merged, thanks!

PR #466, opened 2016-03-17
firmata: working PWM

Propanu:
Nailed it!


Propanu:
Thanks @rwaldron for fixing period values.

In theory all pwm functions should now be usable with subplatforms.

PR #476, opened 2016-03-23
mraajs: Allow user to override node module install path.

pylbert:
Currently, node modules install under the current environment's
node install path.  This happens even if cmake is passed a
CMAKE_INSTALL_PREFIX.

This change checks if a CMAKE_INSTALL_PREFIX has been provided,
in which case it prepens the install path to the NODE_MODULE_INSTALL_PATH.
If a CMAKE_INSTALL_PREFIX has NOT been provided, the the mraa node
modules get install according to the NODE_ROOT_DIR.

This fixes a problem with an install case where the caller does not
have write access to the NODE_ROOT_DIR.

### DIFF
--- a/src/javascript/CMakeLists.txt
+++ b/src/javascript/CMakeLists.txt
@@ -54,7 +54,15 @@ if (${V8_VERSION_MAJOR} GREATER 3)
   endif ()
 endif ()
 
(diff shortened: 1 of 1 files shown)


arfoll:
I like it. Will merge in morning. Thx

arfoll:
Merged. Thanks

PR #486, opened 2016-03-28
LIB_INSTALL_DIR now honored by builds

Propanu:


arfoll:
Merged this mostly - thanks for looking at this!

Propanu:
You grabbed the older commits for this, I changed them a bit yesterday. That IF ELSE in Cmake is superfluous when ` include(GNUInstallDirs)` is called first. I'll submit a patch.

PR #491, opened 2016-04-11
raspberry_pi.c: gpio mmaping over /dev/gpiomem

4refr0nt:
Signed-off-by: Victor Brutskiy <4refr0nt@gmail.com>


arfoll:
I like it, but can we not be cleverer? On mraa_init() inside the rpi can we not just check if /dev/gpiomem exists and if it does use that?

4refr0nt:
Other people may want work with `/dev/mem` only

arfoll:
is there a reason to use /dev/mem when /dev/gpiomem is accesible?

4refr0nt:
I do not know, but it does not mean that no reasons

PR #494, opened 2016-04-22
Add support for android build

robertchiras:
Using NPM, this module can be cross-compiled for Android using NDK.
But, when compiling for Android, we need to add glob.c to source files
(as in Android.mk).
Also, re-arranged the layout in binding.gyp for better readability.

Signed-off-by: Robert Chiras <robert.chiras@intel.com>


arfoll:
Merged, please next time split styling fixes and functional fixes into 2 patches.

PR #503, opened 2016-05-03
ft4222: fixed PCA9555 expander detection

Propanu:
This fixes expander detection when connected to an FTDI4222. A "cold" PCA9555 needs a write before you can begin reading. Subsequent reads will cycle the input register though (0x00, 0x01). Thanks @whbruce for doing a great job merging my functionality otherwise.

Signed-off-by: Propanu <mihai.tudor.panu@intel.com>


arfoll:
Set user.name in git to something real ;-) Corrected for you in this patch & merged.

PR #513, opened 2016-05-20
mraa.c: Added common function to init io from a description

Hbrinj:
mraa_init_io: Added to allow io to be initialised using a description
C++ IO constructors: Added constructors which take a void* to an IO context, used to help wrap the creation of IO using a description.
Aio Constructor: changed the unsigned int constructor to a normal int type, ambiguity errors arose due to the double meaning of unsigned 0 in c++, either the unsigned 0 or a void* to a NULL
mraa.i: Added ignore's for the void* constructors for Aio,Pwm etc., this was because java doesn't know how to handle them, and they also don't need to be in the swig derived languages.

### DIFF
--- a/api/mraa/aio.hpp
+++ b/api/mraa/aio.hpp
@@ -47,11 +47,24 @@ class Aio
      *
      * @param pin channel number to read ADC inputs
      */
-    Aio(unsigned int pin)
+    Aio(int pin)
     {
         m_aio = mraa_aio_init(pin);
         if (m_aio == NULL) {
             throw std::invalid_argument("Invalid AIO pin specified - do you have an ADC?");
+        }
+    }
+    /**
+     * Aio Constructor, takes a pointer to the AIO context and initialises
+     * the AIO class
+     *
+     * @param void * to an AIO context
+     */
+    Aio(void* aio_context)
+    {
+        m_aio = (mraa_aio_context) aio_context;
+        if (m_aio == NULL) {
+            throw std::invalid_argument("Invalid AIO context");
         }
     }
     /**
--- a/src/mraa.c
+++ b/src/mraa.c
@@ -41,6 +41,9 @@
 #include <string.h>
 #include <stdio.h>
 #include <stdbool.h>
+#include <errno.h>
+#include <ctype.h>
+#include <limits.h>
 
 #if defined(IMRAA)
 #include <json-c/json.h>
@@ -52,6 +55,12 @@
 #include "firmata/firmata_mraa.h"
 #include "gpio.h"
 #include "version.h"
+#include "i2c.h"
+#include "pwm.h"
+#include "aio.h"
+#include "spi.h"
+#include "uart.h"
+
 
 #define IIO_DEVICE_WILDCARD "iio:device*"
 mraa_board_t* plat = NULL;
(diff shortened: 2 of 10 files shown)


src/mraa.c
```diff
@@ -1073,3 +1081,153 @@ mraa_add_from_lockfile(const char* imraa_lock_file)
     return ret;
 }
 #endif
+
+void
+mraa_to_upper(char* s)
+{
+    char* t = s;
+    for(;*t;++t)
+        *t = toupper(*t);
```
tingleby: I think even on liners are preferred in {}

Hbrinj: doneski

alext-mkrs:
Commit message makes no sense to me personally, when taken together with the title, I'd suggest to rephrase it.

What's the goal of this one in general?

And there are magic constants everywhere in the helper function, that calls for a correction. The helper itself is long and has no comments whatsoever, adding them and explaining operations being done there would make it much easier to read and comprehend for anyone looking at the code.

Hbrinj:
Thanks for the feedback, i've updated the description to hopefully give a bit more clarity. Can you give me a bit more information regarding which magic constants your talking about?

src/mraa.c
```diff
@@ -1073,3 +1082,131 @@ mraa_add_from_lockfile(const char* imraa_lock_file)
...
+    for (; *t; ++t) {
+        *t = toupper(*t);
+    }
+}
+
+mraa_result_t
+mraa_atoi(char* intStr, int* value)
+{
+    char* end;
+    long val = strtol(intStr, &end, 10);
```
alext-mkrs: magic "10"

src/mraa.c
```diff
@@ -1073,3 +1082,131 @@ mraa_add_from_lockfile(const char* imraa_lock_file)
...
+    return mraa_atoi(token, value);
+}
+
+void*
+mraa_init_io(const char* desc)
+{
+    const char* delim = "-";
+    int length = 0, raw = 0;
+    unsigned int pin = 0, id = 0;
+    char _desc[256] = { 0 }, type[8] = { 0 };
```
alext-mkrs: 256 and 8 are magic constants

src/mraa.c
```diff
@@ -1073,3 +1082,131 @@ mraa_add_from_lockfile(const char* imraa_lock_file)
...
+    length = strlen(desc);
+    if (length > 255 || desc == NULL || length == 0) {
+        return (void*) NULL;
+    }
+    strncpy(_desc, desc, length);
+
+    str = _desc;
+    token = strsep(&str, delim);
+    length = strlen(token);
+    if (length > 4) {
```
alext-mkrs: 4 is magic constant

alext-mkrs:
I've commented on a couple of lines with magic constants (but not all of them, there are more).

With corrected description it now makes more sense, but at the same time it's clear there are four unrelated changes in a single commit, you need to split them into four commits.

Where is that "description" init way is useful, could you provide any examples where that works better than the current way?

api/mraa/aio.hpp
```diff
@@ -47,14 +47,27 @@ class Aio
      *
      * @param pin channel number to read ADC inputs
      */
-    Aio(unsigned int pin)
+    Aio(int pin)
```
tingleby: Not sure this should be bundled into the main commit? @arfoll better in a separate commit? Just in case it breaks something somewhere.

tingleby:
I think it could do with some more logging output when errors occur. Like if the string inst recognised, the mode or whatever.

api/mraa/aio.hpp
arfoll: I dont' want this split out because otherwise the main commit won't work. Having it on it's own doesn't make sense imho because the reason for it is the main commit. Kind of a catch 22.

@Hbrinj I'd recommend you split it like this:
1. mraa_io_init C only
2. C++ template interface
3. Swig stuff on top

src/mraa.c
```diff
@@ -1073,3 +1082,131 @@ mraa_add_from_lockfile(const char* imraa_lock_file)
...
+        }
+        return (void*) mraa_pwm_init(pin);
+    } else if (strncmp(type, "SPI", 3) == 0) {
+        if (raw) {
+            if (mraa_init_io_helper(&str, &id, delim) == MRAA_SUCCESS) {
+                return (void*) NULL;
+            }
+            if (mraa_init_io_helper(&str, &pin, delim) == MRAA_SUCCESS) {
+                return (void*) mraa_spi_init_raw(id, pin);
+            } else {
```
arfoll: Useless else clause?

src/mraa.c
arfoll: You have `desc` and `_desc`, it's pretty confusing...

arfoll:
Merged. Thanks everyone for reviewing & pushing this through. You rock!

Issue #44, opened 2014-11-23
Support installation from NPM for Nodejs


AdamMagaluk:
Currently installing mraa Nodejs requires installing from `opkg`, this is normal for c++ libraries but rarely in the Nodejs community. In the Nodejs community dependencies are handled through the [npm](https://npmjs.org) package management system, which does support native modules.

I would like help with this anyway possible. I currently have a module [mraa-js](https://github.com/AdamMagaluk/mraa-js) that a) deploys the built mraa.node from opkg b) has a set of mocks for development on non edison boards.

Options for native packaging native modules:
1. Add the .node file to the package, since the module is made to support only a number of systems this approach may be optimal.
2. Add a native build step and let node-gyp build mraa when npm install runs.
3. Others??

Any thoughts would be great.
Adam

arfoll:
@AdamMagaluk would you mind taking down the mraa-js module on npm since I don't think it's required anymore and has caused confusion for some users :/

AdamMagaluk:
Will do. Give a day or two gotta update and test some dependencies that rely on it.

jjpr:
@arfoll, did anything ever come of mraa-mock?  I've found that the inability to install mraa for node on a desktop OS makes it difficult for a developer who has an Edison to collaborate with a developer who doesn't have an Edison.  (It's easier to work on a web UI if you run the server, but the server dies on "require('mraa')".)  For now I've repurposed [the mock from galileo-io](https://github.com/rwaldron/galileo-io/blob/master/test/mraa-mock.js), but my efforts seem like a hack.  If you found a more elegant way, it might benefit others.  Thanks.

dantler:
@jjpr I put a write-up for how to do this in a PR, #514 .  See [this file](https://github.com/dantler/mraa/tree/master/jsstub) for usage instructions.  Please let me know if you have any feedback inside #514!

PR #514, opened 2016-05-20
Add JavaScript mraaStub project

dantler:
Signed-off-by: David Antler <david.a.antler@intel.com>

### DIFF
--- a/jsstub/Gruntfile.js
+++ b/jsstub/Gruntfile.js
@@ -1,0 +1,17 @@
+module.exports = function(grunt) {
+  grunt.loadNpmTasks('grunt-mocha-test');
+  // Project configuration.
+  grunt.initConfig({
+    pkg: grunt.file.readJSON('package.json'),
+    mochaTest: {
+      test: {
+        options: {
+          reporter: 'spec'
+        },
+        src: ['test/*.js']
+      }
+    }
+  });
+
+  grunt.registerTask('test', ['mochaTest:test']);
+};
--- a/jsstub/README.md
+++ b/jsstub/README.md
@@ -1,0 +1,51 @@
+mraaStub - JavaScript simulation and stubs for mraa
+====================
+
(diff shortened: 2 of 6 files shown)


dantler:
First PR.  I'm looking for feedback.

Please note that this is very Windows-specific.  Also note that mraaStub has not been added to npm (yet), although `jsstub/README` implies that it has been!

alext-mkrs:
FWIW, that's an interesting idea, though it mixes up low-level functions from mraa domain with actual "device" functionality, which is more of a UPM domain.

And to be honest  I'd rather approach this from a more generic prospective and implement this in mraa itself, so that it's available from all languages mraa supports and not only JS.

This could also solve a problem with someone developing on Windows and needing an IDE autocomplete.

Interesting coincidence with an update in issue #44, it would seem that mocking mraa functionality is suddenly a hot topic :smiley:

dantler:
@alext-mkrs Thanks for taking a look!  The purpose of this library was to allow emulation of a device's specification.  Generally the use case is to write code to target a device that isn't available or may not yet exist.  This has been very useful for me!  

Agreed that mraa is the right place for this item.  It would be nice to support all languages and bindings that mraa supports!  However, I think that task would be complex (you might end up building a real emulator!), and the unit testing culture might be strongest among JavaScript developers anyway.  Has anyone outside of nodejs users requested something like this?

Which autocomplete do you support?  Is there something I could add to support autocomplete?

arfoll:
I've merged this as is (with a slight change in Readme to point to npm as 'future'). We'll try make this API complete and then use it as the basis for the JS docs instead of our current hack. Autocomplete on IDE basically just means that mraa is a loadable & valid node.js lib on windows I don't think anything else is needed.

Thanks!

dantler:
I did put a [mraaStub project up on npm](https://www.npmjs.com/package/mraaStub) and added @arfoll as a collaborator.  Please let me know if you want me to change anything.

Issue #91, opened 2015-01-11
[API] PWM should disable pin when 0 is written


Fleker:
If one writes 0 to a PWM pin, it should automatically disable that pin.

    var pin = new mraa.Pwm(pin, -1, false);
    pin.enable(false) // Should be called if PWM is 0. 

I'm working with some LEDs and they keep showing through even when they have a value of 0. They only stop working when manually disabled.

Also, the Node.js API is poorly documented. I'm not too sure what each part of the constructor is. I'm assuming they are the pin number, `period_us`, and enabled.

arfoll:
Hi @Fleker, try the C++ API docs - http://iotdk.intel.com/docs/master/mraa/classmraa_1_1_pwm.html#a234a0b528c55d2157de74abf4b38daa8 for the argument description.

You should really only be providing the first constructor parameter unless you understand what owner/chipid means. Note those two args will likely get switched around since 'owner' is a prop people will want to set on a non raw pwm object. See #93.

The API docs, I'm working on we'll publish them again soon. I've opened #92.

Let's keep this bug for discussion on just what you first mention - we should disable the pin when writing 0.

miminashi:
I also facing same problem with IntelEdison + mraa 0.9.0 Python binding.
I'm not sure if it is hardware's matter or mraa's matter.
If it is mraa matter, I wish this issue will be solved.

`write(0)` to 3 LEDs, rightmost LED is `enabled(False)`.
![img_5459](https://cloud.githubusercontent.com/assets/90422/12413270/f24788b2-bed0-11e5-804c-33e6d3adbb3e.jpg)

alext-mkrs:
I've started looking into this yesterday and looks like there's a bigger problem with either the pwm IC driver or the IC itself. Every couple of enable/set duty/disable ends with a pin at what seems to be a 100% duty, instead of disabled. At that stage any further enable/disable attempts won't work and only setting other duty cycle gets it out of this state.

I have to look further into this, but if that's an Edison-specific thing, and the one that seems to be broken in several places, maybe we better fix it or do it in a board-specific way rather than changing semantics of duty cycle for everyone.

@arfoll, what do you think?

alext-mkrs:
I've checked the driver with some debug output enabled and this might even be a HW bug/limitation, as the driver reports that it honestly sets the enable/disable bit, but every now and then (but not 100% repeatable) that ends in a 100% duty instead of disabled.

[...]

PR #515, opened 2016-05-24
PWM: added workaround for Edison's problem with 0% duty

alext-mkrs:
Proposed fix (rather a proper workaround, which uses board-specific advance functions) for #91.

Two things to mention aside from the logic itself, which should be rather clear from the code:

1. I haven't run clang-format, as there are quite a few existing lines that are longer than .clang-format allows
1. I haven't split adding `pwm_write_pre()` and `pwm_enable_pre()` into the `mraa_adv_func_t` as while they can be viewed as logically separate changes, splitting them would require always applying commits in a certain order (first adv_func_t, then the workaround), which IMHO is more confusing than helpful.

Let me know if you want me to do either of the above.

There's still a thing, which I mentioned in the triaging notes and not related to #91 - when cycling through `write(<non-zero)/enable(0)` pattern (or equivalents directly from the shell - i.e. this is not a mraa thing at all), sporadically instead of disabling the pin, it would set it to something looking like 100%. I'll open a separate issue for debugging this one.

### DIFF
--- a/src/x86/intel_edison_fab_c.c
+++ b/src/x86/intel_edison_fab_c.c
@@ -76,6 +76,9 @@ static int mmap_fd = 0;
 static int mmap_size;
 static unsigned int mmap_count = 0;
 
+// PWM 0% duty workaround state array
+static int pwm_disabled[4] = { 0 };
+
 mraa_result_t
 mraa_intel_edison_spi_lsbmode_replace(mraa_spi_context dev, mraa_boolean_t lsb)
 {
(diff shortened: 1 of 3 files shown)


src/x86/intel_edison_fab_c.c
```diff
@@ -76,6 +76,9 @@ static int mmap_fd = 0;
 static int mmap_size;
 static unsigned int mmap_count = 0;
 
+// PWM 0% duty workaround state variable
+static int pwm_disabled = 0;
```
arfoll: This is only going to work for one PWM? how are multiple pwm channels handled?

alext-mkrs: Oh well, who uses multiple PWM channels at once or needs more than 640K of RAM anyway :smile: 

Seriously though, I apparently missed this use case and focused on a single-channel one. I'll have this updated in the evening today.

As for the state variable - generally I just wanted to keep the current semantics where one can set duty and period as well as enable/disable PWM independently. Simply disabling it on 0.0f and enabling any time we do a `write()` is of course less code and simpler logic, but i think that might take someone by surprise and run some servos when they don't expect that.

alext-mkrs:
Updated the PR to support multiple pins.

src/x86/intel_edison_fab_c.c
```diff
@@ -414,6 +441,7 @@ mraa_intel_edison_pwm_init_pre(int pin)
 mraa_result_t
 mraa_intel_edison_pwm_init_post(mraa_pwm_context pwm)
 {
+    memset(pwm_disabled, 0, sizeof(pwm_disabled));
```
arfoll: Why is all pwm status being disabled on init? Surely only the pwm disabled flag for the pin in question should be disabled?

alext-mkrs: Argh, true. That lack of time sometimes gets me...
Corrected in the updated commit, also rebased on top of current master.

alext-mkrs:
Anything else you want me to do for this PR? I'm thinking of maybe adding some tests for PWM, but we'd need to define which platforms to target and what hw setup to expect. And then that's not directly the PR's goal, so probably should be a separate one.

alext-mkrs:
@arfoll, ping :smiley:

Anything else you'd want to be done here?

arfoll:
Nope, this is good to go. Merged into master. Thanks alot!

alext-mkrs:
Thanks!

PR #516, opened 2016-05-27
intel_nuc5.c: Fixed incorrect I2C numbering

Hbrinj:
The pin number was being shifted by one instead of 2, resulting in an incorrect mapping. iteration 0 mapped to pins 12 and 13 iteration 1 mapped to 13 and 14 should have been 14 and 15.


arfoll:
Merged, thanks

PR #517, opened 2016-05-27
intel_edison_fab_c.c: fixed the pins not being freed

Hbrinj:
The pins for the Edison Arduino expansion where not being freed in the case where a tristate was not successful


arfoll:
Merged, thanks

PR #519, opened 2016-06-01
mraa.c: Fixed logic error when deciding if the description is null

Hbrinj:


arfoll:
Thanks!

PR #522, opened 2016-06-03
ADD gpio: support to old fashion gpio path names

bmeneguele:
Old kernel version (3.x) use gpio paths in the format of
gpio\<pin_number\>_\<pin_name\>, where both were defined at .fex files
passed by the bootloader to the kernel with hardware information.
Libmraa didn't support this kind of paths because all gpio operations
were done with hard coded path names. With this commit it's possible to
run libmraa either on newer kernel version and older ones, with both
formats of gpio sysfs paths.

Also, some corrections were made in some parts of doxygem doc, just
small things.

Signed-off-by: Bruno E. O. Meneguele <bmeneguele@gmail.com>

### DIFF
--- a/include/mraa_internal_types.h
+++ b/include/mraa_internal_types.h
@@ -59,6 +59,7 @@ struct _gpio {
     int pin; /**< the pin number, as known to the os. */
     int phy_pin; /**< pin passed to clean init. -1 none and raw*/
     int value_fp; /**< the file pointer to the value of the gpio */
+    char *gpio_path; /**< Custom file path format to help with compatibility */
     void (* isr)(void *); /**< the interrupt service request */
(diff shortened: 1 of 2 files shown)


bmeneguele:
In fact, there is a problem in this code that I didn't realize till now. I'll close this PR and open a new one on Monday. Sorry for that, it's a basic error, but I won't have opportunity to test the solution until Moday.

But, I'll be back =).

arfoll:
Ehehe no worries. I'm away all next week so will review week after, @hbrinj
will run the test suite whenever you send your new pr.

Cheers!

alext-mkrs:
Gpio test in the tests dir will have to be updated for this change, as currently the path is hardcoded there to /sys/class/gpio/gpioNUM, i.e. namely the one @bmeneguele wants to change.

PR #523, opened 2016-06-07
Support to old fashion GPIOs class path names

bmeneguele:
Old kernel version (3.x) use gpio paths in the format of gpio<pin_number>_<pin_name>, where both were defined at .fex files passed by the bootloader to the kernel with hardware information.

Libmraa didn't support this kind of paths because all gpio operations were done with hard coded path names. With this commit it's possible to run libmraa either on newer kernel version and older ones, with both formats of gpio sysfs paths.

This PR also guarantees the kernel version discovery either when the gpio was already exported by any other process or when it is the first time used.


alext-mkrs:
Haven't had a chance to analyze it in full yet, but what's immediately apparent is that it makes sense to squash both commits into one. And looks like you've removed a couple of "not owner" comments, which IMHO were helpful.

bmeneguele:
@alext-mkrs done, squashed both commits. But about the comments I removed: I added new comments a little bit more generic just before the 'if' block about this 'ownership', do you think would be better to revert it?

I saw that on Travis CI the commits failed, but the failure report shows a problem to the test script itself:

```
$ sudo apt-get install -y -qq swig3.0 python git cmake

WARNING: The following packages cannot be authenticated!

  cmake cmake-data

E: There are problems and -y was used without --force-yes

The command "sudo apt-get install -y -qq swig3.0 python git cmake" failed and exited with 100 during .
```

src/gpio/gpio.c
```diff
@@ -72,6 +78,7 @@ mraa_gpio_init_internal(mraa_adv_func_t* func_table, int pin)
 
     dev->advance_func = func_table;
     dev->pin = pin;
+    dev->gpio_path = (char*) calloc(MAX_SIZE, sizeof(char));
```
Hbrinj: Won't this need to be freed in mraa deinit

alext-mkrs:
Thanks for squashing, but now you have duplicated Signed-off-by line and two commit messages in one, I'd suggest just rephrase and merge them.

As for the comment - let me take a closer look on a  computer in a day or two, as phone is not an ideal tool :smiley:

bmeneguele:
@Hbrinj well pointed. Indeed it's also needed to check its state after allocation and freed in case of error (goto init_internal_cleanup). Now it's done.

@alext-mkrs sorry for theses commits =), but now it's done. Take a look when you have some time.
Thanks.

include/mraa_internal_types.h
```diff
@@ -215,7 +216,7 @@ typedef struct {
     /*@{*/
     unsigned int pinmap; /**< sysfs pin */
     unsigned int parent_id; /** parent chip id */
-    unsigned int mux_total; /** Numfer of muxes needed for operation of pin */
+    unsigned int mux_total; /** Number of muxes needed for operation of pin */
```
alext-mkrs: Styling/typo fixes not directly related to the change at hand should better be split out into separate commits.

Othserwise they pollute git history and make it harder to git bisect/blame

include/mraa_internal_types.h
```diff
@@ -304,8 +305,8 @@ typedef struct _board_t {
     unsigned int def_spi_bus; /**< Position in array of defult spi bus */
     unsigned int adc_raw; /**< ADC raw bit value */
     unsigned int adc_supported; /**< ADC supported bit value */
-    unsigned int def_uart_dev; /**< Position in array of defult uart */
-    int uart_dev_count; /**< Usable spi Count */
+    unsigned int def_uart_dev; /**< Position in array of default uart */
```
alext-mkrs: Same story with typos

include/mraa_internal_types.h
```diff
@@ -304,8 +305,8 @@ typedef struct _board_t {
...
+    int uart_dev_count; /**< Usable UART Count */
```
alext-mkrs: Same story with typos

src/gpio/gpio.c
```diff
@@ -44,13 +45,13 @@ static mraa_result_t
...
     char bu[MAX_SIZE];
-    sprintf(bu, SYSFS_CLASS_GPIO "/gpio%d/value", dev->pin);
+
+    sprintf(bu, SYSFS_CLASS_GPIO "%s/value", dev->gpio_path);
     dev->value_fp = open(bu, O_RDWR);
     if (dev->value_fp == -1) {
         syslog(LOG_ERR, "gpio%i: Failed to open 'value': %s", dev->pin, strerror(errno));
         return MRAA_ERROR_INVALID_RESOURCE;
     }
-
```
alext-mkrs: Same as with typos styling fixes are better be avoided in functional changes

src/gpio/gpio.c
```diff
@@ -61,68 +62,107 @@ mraa_gpio_init_internal(mraa_adv_func_t* func_table, int pin)
...
+    char bu[MAX_SIZE], pin_name[MRAA_PIN_NAME_SIZE];
+    int export_fd;
+    int length, i;
 
     mraa_gpio_context dev = (mraa_gpio_context) calloc(1, sizeof(struct _gpio));
     if (dev == NULL) {
         syslog(LOG_CRIT, "gpio%i: Failed to allocate memory for context", pin);
         return NULL;
     }
-
```
alext-mkrs: surplus styling fix

src/gpio/gpio.c
```diff
@@ -61,68 +62,107 @@ mraa_gpio_init_internal(mraa_adv_func_t* func_table, int pin)
...
     dev->advance_func = func_table;
     dev->pin = pin;
-
+    dev->gpio_path = (char*) calloc(MAX_SIZE, sizeof(char));
+    if (dev->gpio_path == NULL) {
+        syslog(LOG_CRIT, "gpio%i: Failed to allocate memory for context's attribute", pin);
```
alext-mkrs: it would be helpful to mention which attribute we are talking about here, helps debugging problems

src/gpio/gpio.c
```diff
@@ -61,68 +62,107 @@ mraa_gpio_init_internal(mraa_adv_func_t* func_table, int pin)
...
+        goto init_internal_cleanup;
+    }
     if (IS_FUNC_DEFINED(dev, gpio_init_internal_replace)) {
         status = dev->advance_func->gpio_init_internal_replace(dev, pin);
         if (status == MRAA_SUCCESS)
             return dev;
         else
             goto init_internal_cleanup;
     }
-
```
alext-mkrs: surplus styling fix (BTW doesn't it make it harder to read, by removing whitespace between logical blocks?)

src/gpio/gpio.c
```diff
@@ -61,68 +62,107 @@ mraa_gpio_init_internal(mraa_adv_func_t* func_table, int pin)
...
     if (IS_FUNC_DEFINED(dev, gpio_init_pre)) {
         status = dev->advance_func->gpio_init_pre(pin);
         if (status != MRAA_SUCCESS)
             goto init_internal_cleanup;
     }
-
```
alext-mkrs: surplus styling fix

src/gpio/gpio.c
```diff
@@ -61,68 +62,107 @@ mraa_gpio_init_internal(mraa_adv_func_t* func_table, int pin)
...
     dev->value_fp = -1;
     dev->isr_value_fp = -1;
 #ifndef HAVE_PTHREAD_CANCEL
     dev->isr_control_pipe[0] = dev->isr_control_pipe[1] = -1;
 #endif
     dev->isr_thread_terminating = 0;
     dev->phy_pin = -1;
+    dev->owner = 1;
+    snprintf(dev->gpio_path, MAX_SIZE, "/gpio%d", dev->pin);
+    snprintf(bu, sizeof(bu), SYSFS_CLASS_GPIO "%s", dev->gpio_path);
```
alext-mkrs: It seems like `gpio_path` is empty at this point,just a freshly allocated and zeroed out piece of memory. What's the idea here?

src/gpio/gpio.c
```diff
@@ -44,13 +45,13 @@ static mraa_result_t
 mraa_gpio_get_valfp(mraa_gpio_context dev)
 {
     char bu[MAX_SIZE];
-    sprintf(bu, SYSFS_CLASS_GPIO "/gpio%d/value", dev->pin);
+
+    sprintf(bu, SYSFS_CLASS_GPIO "%s/value", dev->gpio_path);
```
alext-mkrs: IMHO the original spacing was more natural as declaring `bu` and populating it were grouped together. Now populating is grouped with `value_fp`-related code, which is unrelated to `bu`

bmeneguele: I usually don't consider semantics in this situation, I always separate declaration and its use/population, as others coding style standards mention (K&R for instance), but it's personal =P. I don't mind to revert it.

src/gpio/gpio.c
```diff
@@ -61,68 +62,107 @@ mraa_gpio_init_internal(mraa_adv_func_t* func_table, int pin)
...
-        dev->owner = 1;
-        close(export);
+
+        // check for an old or a new kernel version
+        snprintf(dev->gpio_path, MAX_SIZE, "/gpio%d_%s", dev->pin, pin_name);
+        snprintf(bu, sizeof(bu), SYSFS_CLASS_GPIO "%s", dev->gpio_path);
+        if (stat(bu, &dir_stat) == -1 || !S_ISDIR(dir_stat.st_mode))
+            snprintf(dev->gpio_path, MAX_SIZE, "/gpio%d", dev->pin);
+
+        close(export_fd);
```
alext-mkrs: not critical, but we could do this as soon as  line 145 instead of waiting until here.

src/gpio/gpio.c
bmeneguele: Indeed I tried to follow the project coding style, but I didn't realize the "logic/semantic" around it =) I'm sorry for all these coding styles missfix.

src/gpio/gpio.c
```diff
@@ -61,68 +62,107 @@ mraa_gpio_init_internal(mraa_adv_func_t* func_table, int pin)
...
     }
 
 init_internal_cleanup:
     if (status != MRAA_SUCCESS) {
-        if (dev != NULL)
+        if (dev->gpio_path != NULL) {
```
alext-mkrs: `dev` might be NULL here (that's why we are checking for that at line 162), though it's not that likely, so trying to get to `gpio_path` may be dangerous.

To be on the safe side I'd add a check for `dev` being NULL.

bmeneguele:
Indeed, before we continue, I'll close this Pull Request because I made a terrible mistake using the sentence "old kernel version", because this different GPIO class path name is particularity to old kernel versions adopted by the Allwinner SoC vendor, which has its own internal things inside kernel translating the gpios to these schema. I'll change some things and come back later just with the dev->gpio_path generic attribute part, the path schema itself (gpio\<pin_number\>_\<pin_name\>) I'll handle inside Cubietruck's code, with gpio_init_internal_replace(dev, pin).

src/gpio/gpio.c
```diff
@@ -136,7 +176,6 @@ mraa_gpio_init(int pin)
         syslog(LOG_ERR, "gpio%i: init: platform not initialised", pin);
         return NULL;
     }
-
```
alext-mkrs: Surplus styling fix

src/gpio/gpio.c
```diff
@@ -146,7 +185,6 @@ mraa_gpio_init(int pin)
         }
         pin = mraa_get_sub_platform_index(pin);
     }
-
```
alext-mkrs: Surplus styling fix

src/gpio/gpio.c
bmeneguele: Good point!

src/gpio/gpio.c
```diff
@@ -61,68 +62,107 @@ mraa_gpio_init_internal(mraa_adv_func_t* func_table, int pin)
...
             status = MRAA_ERROR_INVALID_RESOURCE;
             goto init_internal_cleanup;
         }
-        dev->owner = 1;
-        close(export);
+
+        // check for an old or a new kernel version
+        snprintf(dev->gpio_path, MAX_SIZE, "/gpio%d_%s", dev->pin, pin_name);
+        snprintf(bu, sizeof(bu), SYSFS_CLASS_GPIO "%s", dev->gpio_path);
+        if (stat(bu, &dir_stat) == -1 || !S_ISDIR(dir_stat.st_mode))
```
alext-mkrs: I think in mraa the recommended style is to enclose even one-line `if` to brackets

bmeneguele: I don't really know, in some places there are both situations one next other, for example in "mraa_gpio_init(int pin)" there are: 

```
    if (r == NULL) {
        return NULL;
    }
    if (r->phy_pin == -1)
        r->phy_pin = pin;
```

I think would be a good idea to do write a small doc about its coding style to help new developers (like me :) ).

alext-mkrs:
FWIW aside from those line comments I've added it looks ok to me by inspection. haven't tried to run or explore it in runtime though.

alext-mkrs:
> <...> I'll change some things and come back later just with the dev->gpio_path generic attribute part, the path schema itself (gpio<pin_number>_<pin_name>) I'll handle inside Cubietruck's code, with gpio_init_internal_replace(dev, pin), and other platforms based on Allwinner might handle the same way. It's a better way, right?

Yes, I think so. If that's a platform-specific quirk, replace functions are there to help that.

> and come back later just with the dev->gpio_path generic attribute part

I wonder if in this case this infrastructure is going to be useful anywhere else and if not - wouldn't using replace functions across the board for other GPIO functionalities (and adding the pin name part in those) be a better solution?

bmeneguele:
> I wonder if in this case this infrastructure is going to be useful anywhere else and if not - wouldn't using replace functions across the board for other GPIO functionalities (and adding the pin name part in those) be a better solution?

Initially I thought it would be the solution, but the hard coded path names are all around others GPIO internal functions that doesn't have replaces. Then I though that the gpio_path attribute on gpio_context_t would help.

alext-mkrs:
I see. You can try adding replaces though :) Similar to what I did with pre- and post- (though replaces would be harder) in #515 

Otherwise you could probably try the split way you mentioned (adding only path and the rest is in replace) and see what @arfoll has to say about that.

src/gpio/gpio.c
alext-mkrs: No worries, that's what PRs are for :smiley:

src/gpio/gpio.c
alext-mkrs: Generally there's a [clang-format](https://github.com/intel-iot-devkit/mraa/blob/master/.clang-format) configuration, that's the ultimate one. Not sure though it covers this particular piece. And it was introduced much later that base of the code, so running it on a file as a whole would not be a good idea :smiley:

However checking out what would it do with your changed/added code and formatting it like that could be a good guidance.

bmeneguele:
I'll try to use the gpio_path first, it's almost done :D, just change little things to clear the "Allwinner" specific parts. But if @arfoll prefers the replace functions, won't be a problem too. But thank you very much @alext-mkrs .. I'll be back soon.

src/gpio/gpio.c
arfoll: Ifs should be enclosed in braces but yes code styling is something I've been a bit lax with, I keep meaning to make commit hooks to just be done with it... another time. Clang-format also is kinda buggy and imho I'm not sure it's the best tool for the job....

arfoll:
I think that by just concatenating either the pin num or the pin path we should be able to get both working relatively easily. I can't review well on my phone but I'll be back next week with a closer look :)

src/gpio/gpio.c
arfoll: If you do pure styling fixes keep them in a separate commit it makes reading much easier. Tbh when merging I try remove such things ;) in this case I'd say leave it unless u fix all the codebase to do that. If you're interested let's open a new issue on what and how we should fix and let's work out something

bmeneguele:
Well, I'm back with new changes, but now I didn't do any style fixes and everything specific to the gpio\<pin_number\>_\<pin_name\> were removed, because as I said before it was a specific case to Allwinner SoC based platforms using old kernel versions. The only thing I made in this commit was to add the gpio_path attribute to struct _gpio and handled it on gpio.c/.h and on the APIs to other languages.

Everything specific to Allwinner SoC must be handled in _replace functions on each platform support, as it's being made in the new support to Cubietruck I'm developing.

api/mraa/gpio.h
```diff
@@ -220,6 +220,14 @@ int mraa_gpio_get_pin(mraa_gpio_context dev);
  */
 int mraa_gpio_get_pin_raw(mraa_gpio_context dev);
 
+/**
+ * Get the SYSFS pin path, invalid will return NULL
+ *
+ * @param dev The Gpio context
+ * @return Gpio pin sysfs path
```
alext-mkrs: I'd suggest to add "or NULL if error", to make it clear.

api/mraa/gpio.hpp
```diff
@@ -315,6 +315,16 @@ class Gpio
         }
         return mraa_gpio_get_pin(m_gpio);
     }
+    /**
+     * Get pin SYSFS path.
+     *
+     * @return Pin path string
```
alext-mkrs: I'd suggest to add "or NULL if error", to make it clear.

src/gpio/gpio.c
```diff
@@ -73,6 +75,14 @@ mraa_gpio_init_internal(mraa_adv_func_t* func_table, int pin)
     dev->advance_func = func_table;
     dev->pin = pin;
 
+    // GPIO sysfs path may store the default path names schema or any other specific to vendor
+    dev->gpio_path = (char*) calloc(MAX_SIZE, sizeof(char));
+    if (dev->gpio_path == NULL) {
+        syslog(LOG_CRIT, "gpio%i: Failed to allocate memory for context's attribute", pin);
```
alext-mkrs: I'd still recommend to elaborate on "context's attribute" and just state directly it's the gpio_path we are talking about.

src/gpio/gpio.c
```diff
@@ -95,34 +105,40 @@ mraa_gpio_init_internal(mraa_adv_func_t* func_table, int pin)
...
+    snprintf(dev->gpio_path, MAX_SIZE, "/gpio%d", dev->pin);
+    snprintf(bu, sizeof(bu), SYSFS_CLASS_GPIO "%s", dev->gpio_path);
+
+    // check if the pin was already created, meaning another process may be the owner
+    if (stat(bu, &dir_stat) == 0 && S_ISDIR(dir_stat.st_mode)) {
+        dev->owner = 0;
     } else {
-        int export = open(SYSFS_CLASS_GPIO "/export", O_WRONLY);
-        if (export == -1) {
+        export_fd = open(SYSFS_CLASS_GPIO "/export", O_WRONLY);
```
alext-mkrs: that `export_fd` is a styling fix, isn't it?

alext-mkrs:
In addition to those inline comments - what's your idea about `mraa_gpio_get_valfp()`, where you haven't updated the `/gpio%d/value` piece? While it's used only behind `_replace()` checks in this file, I think it still would make sense to update it to use `gpio_path` for consistency in case there's a situation where there's no need for `_replace()`, but `gpio_path` is custom. And in case it's not custom `gpio_path` is `gpio%d` anyway, so we have backwards compatibility. What do you think?

api/mraa/gpio.hpp
```diff
@@ -315,6 +315,16 @@ class Gpio
...
+     */
+    char*
+    getPath()
```
arfoll: in C++ this should return a std::string

arfoll:
@bmeneguele can you push an example of a platform that works with this stuff as well? this seems to require a mix of _replace and gpio_path rather than just all moving to gpio_path. I'd much rather simply use gpio_path everywhere and have a small replace in the _init() that just sets gpio_path differently for platforms with names and by default set it to "gpio+dev->pin". Does that makes sense?

src/gpio/gpio.c
bmeneguele: I didn't understand exactly what you mean by "context's attribute", could you give me an example?

api/mraa/gpio.hpp
bmeneguele: well pointed.

bmeneguele:
@alext-mkrs You are right.. I just forgot that peace, there should have the gpio_path as well.

@arfoll Yeah, I will open an PR to the cubietruck support I am working on. It isn't complete yet, I'm facing some SPI problems, but GPIO is working correctly.

api/mraa/gpio.h
bmeneguele: but it already has "invalid will return NULL" in the same way other functions mention.

bmeneguele:
@arfoll take a look at https://github.com/bmeneguele/mraa/commit/e93e261285586d4ec04c4758c541e93760cf8a7d this is the initial support to the Cubietruck board I'm working on right now. I won't pull request yet because I need to change the coding style to the standard of the project, but take a look at line 139 on src/arm/cubietruck.c, that is the _replace function to the gpio_init_internal function that handles the dev->gpio_path.

api/mraa/gpio.h
alext-mkrs: You refer to the comment line and I refer to the `@return` one. Adding it to `@return` will have it fully picked up by the doxygen, meaning proper/better documenting.

src/gpio/gpio.c
alext-mkrs: What I mean is that instead of `"gpio%i: Failed to allocate memory for context's attribute"` make it a `"gpio%i: Failed to allocate memory for gpio_path attribute"` to make it clear, which exact piece were we not able to initialize.

bmeneguele: Wow... I'm so bad on this xD. I'm sorry, you are completely right!

api/mraa/gpio.h
bmeneguele: Oh yeah, I understand and agree. =)

PR #526, opened 2016-06-17
Initialise an unknown platform using a json file

Hbrinj:
Mraa.c: Added conversion from json to mraa_board_t
mraa_internal_types.h: Removed unneeded const, changes unsigned int's to ints
types.h: Added MRAA_JSON_PLATFORM
CMakelists: Added ability to exclude json platform from the build
travisl.yml: Added test for json plaform


api/mraa/types.h
```diff
@@ -50,6 +50,7 @@ typedef enum {
     MRAA_INTEL_SOFIA_3GR = 10,      /**< The Intel SoFIA 3GR */
     MRAA_INTEL_CHERRYHILLS = 11,     /**< The Intel Braswell Cherryhills */
     MRAA_UP = 12,                    /**< The UP Board */
+    MRAA_JSON_PLATFORM = 13,
```
tingleby: Consider moving to a special number 97? @arfoll

src/mraa.c
```diff
@@ -61,8 +62,44 @@
...
+#define PIN_KEY "pin"
+
+#define IO_KEY "layout"
+#define GPIO_KEY "GPIO"
+#define SPI_KEY "SPI"
+#define UART_KEY "UART"
+#define I2C_KEY "I2C"
+#define PWM_KEY "PWM"
+#define AIO_KEY "AIO"
+#define PLATFORM_KEY "platform"
```
tingleby: These defines could be in a internal header

api/mraa/types.h
arfoll: Yeah please move to something special

src/mraa.c
```diff
@@ -1233,3 +1288,668 @@ mraa_init_io(const char* desc)
     syslog(LOG_ERR, "mraa_init_io: Invalid IO type given.");
     return NULL;
 }
+
```
tingleby: The implemented functions for JSON platform could be in its own file "mraa_json.c" or something. Maybe if you think its worth it, its own directory.

Hbrinj: Yeah they can be, i wasn't too sure what the policy is.

tingleby:
Perhaps it would be nice to include an example JSON in tree.

whbruce:
Nice idea. A few comments

- I'm not sure of expected usage, does client call mraa_init_json_platform()? 
- I'd like to see a feature where location of JSON definition file is set a build time (or even by environment variable) and picked up by mraa in place of platform auto-detect.
- As JSON define platform feature is optional move source into a different file so it is not unconditionally compiled
- Add index key to layout schema so we can track pin assignments more easily

I look forward to using this once it's merged.

Hbrinj:
So to answer your second comment, I am going to be adding an environment variable check to load a platform using the value in the env. So the expected usage was either using the env variable or loading mraa and loading the platform. Can explain your last comment further?

api/mraa/common.h
```diff
@@ -44,6 +44,44 @@
...
+#define IO_KEY "layout"
+#define GPIO_KEY "GPIO"
+#define SPI_KEY "SPI"
+#define UART_KEY "UART"
+#define I2C_KEY "I2C"
+#define PWM_KEY "PWM"
+#define AIO_KEY "AIO"
+#define PLATFORM_KEY "platform"
+
+#define JSONPLAT_ENV_VAR "MRAA_JSON_PLATFORM"
```
arfoll: move as much as possible to internal_types.h

api/mraa/common.h
```diff
@@ -306,6 +344,16 @@ mraa_result_t mraa_remove_subplatform(mraa_platform_t subplatformtype);
  */
 void* mraa_init_io(const char* desc);
 
+/**
+ * Instantiate an unknown board using a json file
+ *
+ * @param   Path to the json file, relative to the folder the program 
+ *          was initially run in or a direct path
+ *
+ * @return mraa_result indicating success
```
arfoll: make indentation like the others

examples/platform/turbotjson.json
```diff
@@ -0,0 +1,73 @@
...
+	"pin_count": 28,
+	"gpio_count": 28,
+	"i2c_count": 1,
+	"aio_count": 0,
+	"spi_count": 1,
+	"uart_count": 1,
+	"pwmDefPeriod":500, 
+	"pwmMaxPeriod":1000000000, 
+	"pwmMinPeriod":1
+    }
```
tingleby: little bit of whitespace here

src/CMakeLists.txt
```diff
@@ -120,11 +120,24 @@ if (USBPLAT)
...
-  include_directories (${JSON-C_INCLUDE_DIR})
-  set (mraa_LIBS ${mraa_LIBS} ${JSON-C_LIBRARIES})
+if (JSONPLAT OR IMRAA)
+  find_package (JSON-C QUIET)
+  if (NOT ${JSON-C_FOUND})
+    if (IMRAA)
+      message (FATAL_ERROR "Enabled IMRAA but JSON-C library not found")
+    elseif (JSONPLAT)
+      message (WARNING "Enabled JSONPLAT but JSON-C library not found: Ignoring JSONPLAT")
+    endif ()
```
tingleby: whitespace

src/mraa.c
```diff
@@ -116,6 +123,31 @@ imraa_init()
            mraa_get_version(), (proc_user != NULL) ? proc_user->pw_name : "<unknown>", proc_euid);
 
     mraa_platform_t platform_type;
+
+    // Check to see if the enviroment variable has been set
+    env_var = getenv(JSONPLAT_ENV_VAR);
+    if (env_var != NULL) {
+        // Check to see if it's a file
+        if (access(env_var, R_OK) != -1) {
```
arfoll: asking for permission is pointless since you don't know file changed/detelted etc...

src/mraa.c
```diff
@@ -116,6 +123,31 @@ imraa_init()
            mraa_get_version(), (proc_user != NULL) ? proc_user->pw_name : "<unknown>", proc_euid);
 
     mraa_platform_t platform_type;
+
+    // Check to see if the enviroment variable has been set
+    env_var = getenv(JSONPLAT_ENV_VAR);
```
tingleby: Not sure about this choice. something prepended with MRAA_ would be safer @arfoll what do you think

src/mraa.c
```diff
@@ -116,6 +123,31 @@ imraa_init()
...
+    if (env_var != NULL) {
+        // Check to see if it's a file
```
tingleby: Perhaps for some security we should check its a regular file, or something.

src/mraa.c
```diff
@@ -126,6 +158,7 @@ imraa_init()
 #error mraa_ARCH NOTHING
 #endif
 
+jsonloaded:
```
arfoll: let's replace this by a if (platform_type == MRAA_NULL_PLATFORM) and set platform_type by default to NULL

src/mraa.c
```diff
@@ -1234,3 +1281,687 @@ mraa_init_io(const char* desc)
...
+    free(board->platform_name);
+    free(board->pins);
+    free(board->adv_func);
+    free(board);
+cleanup:
+    json_object_put(jobj_platform);
+    munmap(buffer, st.st_size);
+    return ret;
+#endif
+    return MRAA_ERROR_INVALID_PARAMETER;
```
arfoll: MRAA_ERROR_FEATURE_NOT_SUPPORTED

api/mraa/common.hpp
```diff
@@ -327,4 +327,18 @@ initIo(std::string desc)
     return new T(mraa_init_io(desc.c_str()));
 }
 
+/**
+ * Instantiate an unknown board using a json file
+ *
+ * @param   Path to the json file, relative to the folder the program 
+ *          was initially run in or a direct path
+ *
+ * @return mraa_result indicating success
```
arfoll: Copy comment from another method

src/mraa.c
arfoll: It is, check what that define means. But yeah define could be clearer

src/json/jsonplatform.c
```diff
@@ -0,0 +1,728 @@
...
+ * MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+ * NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE
+ * LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
+ * OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION
+ * WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ */
+
+#define _GNU_SOURCE
+#if !defined(_XOPEN_SOURCE) || _XOPEN_SOURCE < 600
+#define _XOPEN_SOURCE 600 /* Get nftw() and S_IFSOCK declarations */
```
arfoll: but you don't use nftw?

src/json/jsonplatform.c
```diff
@@ -0,0 +1,728 @@
...
+
+#if defined(JSONPLAT)
+#include <json-c/json.h>
+#include <sys/stat.h>
+#include <sys/mman.h>
+#endif
+
+#include "mraa_internal.h"
+
+#if defined(JSONPLAT)
```
arfoll: this file is only included if JSONPLAT so let's not second guess

src/json/jsonplatform.c
```diff
@@ -0,0 +1,728 @@
...
+            return MRAA_ERROR_INVALID_RESOURCE;
+        }
+        *pin = (unsigned int) json_object_get_int(jobj_temp);
+        return MRAA_SUCCESS;
+    }
+    syslog(LOG_ERR, "init_json_platform: No %s specified for %s at position: %d", key, io, index);
+    return MRAA_ERROR_NO_DATA_AVAILABLE;
+}
+
+inline mraa_result_t
```
arfoll: Needs to be inline static if you want to do that. Because we already have -finline-small-functions let's just not make these inline at all and let the compiler work out the best option.

jamietanna:
Looks like you may benefit from #541 for getting Travis on Trusty.

arfoll:
Merged

PR #532, opened 2016-06-26
x86.c: fixed typo in Tuchuck board forced enablement

alext-mkrs:
Looks like a typo and it's not going to work without that.


arfoll:
Thanks alot for this patch. merged.

PR #533, opened 2016-06-26
cmake: fixed incorrect MRAAPLATFORMFORCE definition

alext-mkrs:
This looks like a typo as `ALL` is apparently a default value we want it to have and right now it's a docstring.

It worked all the time due to the fact that during the actual usage we check for "if != 'ALL' <...> else <...>', so it treats everything, which is not a specific platform name as "ALL", but I think it is still worth fixing.


arfoll:
100% correct, thanks for fixing. Merged.

PR #535, opened 2016-06-27
iio: mount matrix compliant to IIO core kernel standard

laykuanloon:
IIO mount matrix is supported in kernel version 4.6.
Change mraa_iio_get_mount_matrix to take in one more parameter
‘mount matrix sysfs entry name’, because different sensor driver
may expose different mount matrix sysfs entry.

Following link explain the mount matrix usage and sysfs entry:
https://www.kernel.org/doc/Documentation/ABI/testing/sysfs-bus-iio

Signed-off-by: Lay, Kuan Loon <kuan.loon.lay@intel.com>


arfoll:
Merged, thanks alot!

Issue #531, opened 2016-06-26
[RFC] Adding mock capability


alext-mkrs:
Following discussions in #44 and #514 I'm going to dive a bit deeper into this and introduce mocking capability on a more basic level, which would make it available through all language bindings we support, without making extensive changes in the established code base.

This RFC is for the general architectural approach to this, as I can see at least a couple. Please let me know, what do you think, @arfoll, @tingleby and anyone else interested in the matter. Below are my thoughts based on my understanding of mraa's internals and a sort of a vision I have for this mocking capability. If you see any other general way of implementing this, I'm all ears.

1. As long as we don't have a full-blown HAL in the library, where we could intercept calls to the HW and mock real platforms (like Edison), implementing something like this would require hacks and `#ifdef`'s all over the place, so I don't think that would be a good way to go. If we were to go there, we'd probably want to `#ifdef` pieces of current HW-dependent code in our generic (like `gpio.c`) and board-specific (like `intel_nuc5.c`) files.
2. In view of that, more viable approach I see here would be to add this capability as yet another "platform", like if it's an actual piece of HW. This contains the code (as we are using our established infra) and should require only a set of `_replace` functions in addition to the standard platform definition. The drawback is that we are not able to trick the user they're dealing with an actual Edison board, but the positive is that we have cleaner code and faster implementation of the capability + we can define it however we want in terms of pin numbers and their layout.

alext-mkrs:
Just to make it clear - under "platform" I mean either the whole "platform type", like `arm` and `x86` we currently have, or just another "platform" like NUC or Edison, within `x86` type.

The former is cleaner and IMHO logical, the latter would also work and save us some code to copy.

alext-mkrs:
You can see what I'm aiming at in [this branch](https://github.com/alext-mkrs/mraa/tree/mraa-mock) of my mraa repo fork.

alext-mkrs:
As the first commit based on the proposed approach went in, this RFC is no longer needed. Closing.

Issue #375, opened 2015-12-02
Mraa autocomplete Python


BTWS:
Is it possible to get autocompletion in Python an IDE?

I tried PyCham and iPython without luck.

arfoll:
@tingleby didn't you try this?

tingleby:
No, I haven't. But something worth looking at.

alext-mkrs:
Looks like this might be just a matter of [properly configuring PyCharm](http://stackoverflow.com/questions/5049842/autocomplete-in-pycharm-for-python-compiled-extensions?rq=1), @BrechtW, check this out.

BrechtW:
I tried to remove and add the python interpreter. I copied the mraa folder in c:\Python27\Lib\site-packages and did Project Interpreter>Gear Icon>More>Show path for selected interpreter>Refresh. The autocomplete still doesn't work. I still get "No module named mraa".

alext-mkrs:
Ah, looks like you're doing this on Windows, that's not going to work. Python binding consists of a simple .py wrapper, which imports a compiled binary (shared object, .so), which is compiled out of SWIG-generated code and is OS-specific, with Windows not being supported (mraa uses Linux-specific mechanisms). And autocomplete needs to actually import it to work, AFAIU.

BrechtW:
Is it possible to make a sort of list of functions and properties and let PyCharm derive autocomplete of this?

alext-mkrs:
It would look like it's not possible. Both in PyCharm and IPython autocompletion works by actually importing Python modules and indexing/analyzing them, not by using some additional dictionary file or something like this. So when you have such a "Windows IDE - Linux-only module" situation it's beyond of their capabilities.

Just as a wild idea you could try running indexing in PyCharm on Linux and then moving whatever cache it generates to Windows, however that's really a rude hack and I don't believe it's going to work.

alext-mkrs:
Unless anyone has any other ideas, I think we just have to state syntax autocompletion is not supported for Windows-based IDEs due to (a) the fact that mraa is a Linux-only library and (b) the way these IDEs work with autocompletion - by trying to actually import the module and retrospect it.

@arfoll, @tingleby, what do you think?

alext-mkrs:
As mentioned previously, the way autocomplete works is simply not appropriate for a situation when you have a Linux-compiled Python module and Windows-hosted IDE, so I don't think we can make it work. Closing this one.

PR #538, opened 2016-07-02
mraa mock

alext-mkrs:
Here comes the mocking capability foundation + GPIO functionality implementation (without ISR) discussed in #531, #514 and #44 and done the way I deem the best (at the backend and as a separate platform - see my reasoning in #531).

The second commit enables mraa to build under MSYS2 in Windows and the mock platform was tested to work fine there. If nothing else, it solves the problem of autocompletion in Pythin IDEs, raised in #375 :smiley: I've personally tested it to work just fine under PyCharm, see screenshot below.

Your comments and suggestions are more than welcome. There are several things to note:

1. The way we used `BUILDARCH` in CMake, seems to be incorrect - we define it as "option" (i.e. as a bool), but then assign strings to it - so I fixed this to be a string from the beginning. I haven't split it out as a separate commit.
1. I've used 96 as the platform ID as I recall there was a suggestion to use 97 for something else recently.
1. To avoid cutting out the whole `spi.c` file under Windows I had to extract definitions from `spidev.h` to `spi_kernel_headers.h` in the way it was previously done for IIO and for files under `include/linux`. MSYS2 does not have `spidev.h` available at all.
1. I have tried to compile it using bare MinGW, but we rely on POSIX things a bit too much (e.g. `syslog()` all over the place), so it didn't work and would require quite some effort to enable it. MSYS2 on the other hand only adds one DLL dependency, which should be ok, given the tasks mraa mock on Windows is aimed to solve. As a bonus, MSYS2 provides nice complete UNIX-like environment and has a convenient package manager.

And here's mraa in all its glory running under Windows :sunglasses:

pycharm_win_autocompletion


api/mraa/spi_kernel_headers.h
```diff
@@ -0,0 +1,58 @@
...
+#define SPI_IOC_RD_BITS_PER_WORD        _IOR(SPI_IOC_MAGIC, 3, __u8)
+#define SPI_IOC_WR_BITS_PER_WORD        _IOW(SPI_IOC_MAGIC, 3, __u8)
+
+#define SPI_IOC_RD_MAX_SPEED_HZ         _IOR(SPI_IOC_MAGIC, 4, __u32)
+#define SPI_IOC_WR_MAX_SPEED_HZ         _IOW(SPI_IOC_MAGIC, 4, __u32)
+
+#define SPI_IOC_RD_MODE32               _IOR(SPI_IOC_MAGIC, 5, __u32)
+#define SPI_IOC_WR_MODE32               _IOW(SPI_IOC_MAGIC, 5, __u32)
+
+#endif
```
arfoll: Can we have this file in include/ and not in the public API?

arfoll:
I love this thanks @alext-mkrs! I've merged the first patch, the second patch just watining if I can move that header into include/ to make things a bit cleaner.

api/mraa/spi_kernel_headers.h
alext-mkrs: Sure thing, can do that easily - which location do you think would be better: `include/` of `include/linux`?

alext-mkrs:
Glad to read that :smiley:. If you mean you can just move that during the merge, then by all means please do, I picked the location by using the iio include as the example, but I indeed spent some time thinking if `include` or `include/linux` is not a better one.

api/mraa/spi_kernel_headers.h
arfoll: yeah include/linux is perfect for this

arfoll:
ok I'll move it when I merge it no problems. yeah the iio is kept in there because we use the channel types and iio_event_data and I dind't want to uselessly redefine all that stuff. The example to follow is probably the i2c.h one.

alext-mkrs:
Got it. Ok, I have this change on my branch, but haven't pushed - if you want me to do that (including rebasing to the latest master), please just tell.

arfoll:
Well If you've done it then yes go ahead :)

alext-mkrs:
Done

arfoll:
Thanks again! Merged.

PR #543, opened 2016-07-10
Mock: added AIO functionality

alext-mkrs:
Here comes AIO for mraa mock. It could be more elaborated, with additional functions to allow the user to set their own static value to be returned all the time, but I decided to go with a simple random for starters, if anyone needs static, it could be added later.

Commit with `mraa_aio_close()` additions changed enough of the function for a couple of additonal styling fixes (one-line `if` without braces and `return` with value in parentheses - the only place in a whole file we do such things) to become invisible, so I went for that.

### DIFF
--- a/docs/mock.md
+++ b/docs/mock.md
@@ -12,17 +12,17 @@ Board configuration
 -------------------
 
 This feature is yet in the experimental mode and not all functionality is available.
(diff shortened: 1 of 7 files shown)


arfoll:
Perfect. Merged

PR #548, opened 2016-07-21
mraa mock: I2C functionality implementation

alext-mkrs:
This one contains I2C functionality implementation. As usual I tried to keep it simple, but functional, and implemented a no-frills I2C abstract device with 10 bytes of register space. `mraa-i2c` works just fine :smiley: 

I've also added standard context validity checks into `i2c.c`, which it was almost completely lacking. I've also refactored mock implementation a bit and split out particular functionality implementations into separate files to make it more manageable (`mock_board.c` was getting quite big).

Doing that I've spotted something akin to a bug where we had `i2c_stop_replace` defined in `mraa_adv_func.h`, but we actually never called it in `i2c.c`. Fixed that as well and this should probably help some FT4222 memory leaks.


arfoll:
Great stuff, merged!

PR #553, opened 2016-08-01
gpio.h: remove extraneous extern C declaration

jontrulson:

### DIFF
--- a/api/mraa/gpio.h
+++ b/api/mraa/gpio.h
@@ -46,9 +46,7 @@ extern "C" {
 
 #if defined(SWIGJAVA) || defined(JAVACALLBACK)
 #include <jni.h>
-extern "C" {
-    void mraa_java_isr_callback(void *args);
-}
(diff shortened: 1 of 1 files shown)


jontrulson:
Is there anything else that needs to be done to get this fix committed?  It causes build problems with C based drivers.

arfoll:
Not really, I'll take a good look at a bunch of PRS and merge tmrw. Sorry and thx for the poke

jontrulson:
Thanks!

arfoll:
Merged - would be interested in seeing what the issues with this are - can you link me to what failures have been happening?

jontrulson:
Since the clause was not protected by an "#ifdef __cplusplus", any C code compiled with JAVACALLBACK or SWIGJAVA defined would fail, as 'extern "C" ...' isn't valid C.

Since the file is already properly protected via this guard at the top and bottom, this clause was not necessary anyway.

PR #561, opened 2016-08-12
gt: add stubs for i2c_frequency

Propanu:


arfoll:
As discussed via email, I'm not adding this. It would need to be on all platforms which is just really pointless.

PR #564, opened 2016-08-17
mraa.c: Replaced string values with #defines

Hbrinj:
Signed-off-by: Houman Brinjcargorabi <hbrinjcar@gmail.com>

### DIFF
--- a/src/mraa.c
+++ b/src/mraa.c
@@ -1210,7 +1210,7 @@ mraa_init_io(const char* desc)
         return NULL;
     }
 
-    if (strncmp(type, "GPIO", 4) == 0) {
+    if (strncmp(type, GPIO_KEY, strlen(GPIO_KEY)) == 0) {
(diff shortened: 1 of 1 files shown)


src/mraa.c
```diff
@@ -1210,7 +1210,7 @@ mraa_init_io(const char* desc)
         return NULL;
     }
 
-    if (strncmp(type, "GPIO", 4) == 0) {
+    if (strncmp(type, GPIO_KEY, 4) == 0) {
```
alext-mkrs: I'd suggest replacing magic "4" with an automatically computed length, for all these comparisons.

tingleby: sizeof(GPIO_KEY) should be replace by the compiler with a value.Not sure how well this works.

Hbrinj:
Done

arfoll:
Merged thanks!

PR #565, opened 2016-08-18
SPI functionality for mraa mock

alext-mkrs:
One specific note - when adding standard context validity checks to SPI init functions in `spi.c` I had to rework the `init_raw()` function a bit, in a way similar to `init_internal()` in `i2c.c`. Now there's a single cleanup section and all other ones instead of repeating the `free()/return` pattern just redirect there.

I've also noticed there is a couple of surplus includes in the I2C mock implementation, so I've removed them in a separate commit piggybacked to this PR.

### DIFF
--- a/docs/mock.md
+++ b/docs/mock.md
@@ -19,6 +19,10 @@ Right now we simulate a single generic board with:
 which can be read or written in bytes or words (big-endian). Technically those
 registers are just an array of `uint8_t`, so you can treat them as 10 single-byte
 registers or 5 single-word ones or a mix thereof.
+* Single SPI bus with one Chip Select. All write functions are supported, received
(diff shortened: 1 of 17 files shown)


arfoll:
This is awesome - Merged. Could you add a .travis job to build the mock tests and run them? Also only some of the files in tests/mock have +x on them, I've given it to all the files in tests/ I don't really have a preference either way but if you do then shout and let's make them the same :)

alext-mkrs:
Thanks. Sure, I'll prepare a Travis config (was thinking about that actually when waiting for the verification run to finish :smiley:). As for the execute bit - no preference, so having it on for all, as you did, is absolutely fine.

PR #575, opened 2016-09-08
python: remove temporary RPATH from _mraa.so when installing

Propanu:


arfoll:
rebased a bit and merged, thanks!

PR #576, opened 2016-09-09
uart_ow: convert static const int MRAA_UART_OW_ROMCODE_SIZE to #define

jontrulson:
This is needed so that this value can be used in all places in C (and
C++) code.  Unlike C++ you can't always use a static const int
variable to declare an array size for example.  Trying to do so will
yield compiler errors.

In addition, during the declaration of the uart_ow_context, replace a
hardcoded '8' with this macro.  Originally, the '8' was hardcoded for
this very reason.

### DIFF
--- a/api/mraa/uart_ow.h
+++ b/api/mraa/uart_ow.h
@@ -64,18 +64,18 @@ extern "C" {
 #include "common.h"
 #include "uart.h"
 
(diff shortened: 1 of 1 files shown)


alext-mkrs:
The initial build failure was not related to the PR itself, so I reran it.

arfoll:
Merged, thanks!

PR #580, opened 2016-09-14
java: Renamed pom.xml file to the standard naming convention

stefan-andritoiu:


arfoll:
Merged

PR #584, opened 2016-09-27
add support for IOT2000

SaschaWeisenberger:
The Siemens device "SIMATIC IOT2000" is identical to the Intel Galileo
Gen 2 regarding the gpios. So if the board name is "SIMATIC IOT2000"
treat the board as if it is Galileo Gen 2.

Signed-off-by: Sascha Weisenberger <sascha.weisenberger@siemens.com>


arfoll:
Merged, thanks alot

Issue #585, opened 2016-09-28
Setting PWMs on Edison to zero duty is weird and unreliable


RiccardoMaffei:
I'm currently using mraa v1.3.0 with Java bindings on an Intel Edison and sometimes the PWM pins are not set to zero.

In this version the #515 fix should be already implemented.
This bug may be similar and related to the #91 but I think it's not the same.

PWM pins sometimes (**not always**) fail to go to 0% duty but the failure rate seem related to the difference between the zero and the previous duty value. 
I've made some tests including the three examples provided below.
## Example 1:

The following code has a failure rate of ~100%. A led connected to the pin is almost never turned off.

``` java
for(int i = 0; i < 100; i++) {  //a hundred times
    Thread.sleep(500);          //wait half a second
    pwmPin.write(1);            //write 100%
    Thread.sleep(500);          //wait half a second
    pwmPin.write(0.0f);         //write 0%
}
```
## Example 2:

The following code has a failure rate of ~60%:
- ~40% of the time the led works as expected: 100%->50%->0% (each loop)
- ~60% of the time the led is not turned off and goes to 100%: 100%->50%->100% (each loop)

``` java
for(int i = 0; i < 100; i++) {  //a hundred times
    Thread.sleep(500);          //wait half a second
    pwmPin.write(1);            //write 100%
    Thread.sleep(500);          //wait half a second
    pwmPin.write(0.5f);         //write 50%
    Thread.sleep(500);          //wait half a second
    pwmPin.write(0.0f);         //write 0%
}
```
## Example 3:

In the following code the two pins fail but not necessarily at the same time:
- sometimes the first pin fails
- sometimes the second pin fails
- sometimes they both fail
- sometimes they both work as expected

``` java
for(int i = 0; i < 100; i++) {  //a hundred times
    Thread.sleep(500);          //wait half a second
    pwmPinOne.write(1);         //write 100%
    pwmPinTwo.write(1);         //write 100%
    Thread.sleep(500);          //wait half a second
    pwmPinOne.write(0.5f);      //write 50%
    pwmPinTwo.write(0.5f);      //write 50%
    Thread.sleep(500);          //wait half a second
    pwmPinOne.write(0.0f);      //write 0%
    pwmPinTwo.write(0.0f);      //write 0%
}
```

I've tried using `pwmPin.enable(false)` but the bug is still there and the failure rate is the same.

alext-mkrs:
[...]

PR #586, opened 2016-10-04
intel_edison_fab_c.c: Fix edison disabled pwm stuck at high

belugon:
Edison pwm can stuck at high if pin is disabled during ON pwm peroid. #585
Workaround to force zero duty time before disabling the pin.


alext-mkrs:
I see you are changing the enable/disable logic by removing array updates, I'd be careful with that. Will test your proposal and comment later.

You need to sign-off your commit, see contributing.md

src/x86/intel_edison_fab_c.c
```diff
@@ -373,8 +374,17 @@ mraa_intel_edison_aio_init_post(mraa_aio_context dev)
...
     // PWM 0% duty workaround: update state array
-    // if someone first ran write(0) and then enable(1).
-    if ((pwm_disabled[dev->pin] == 1) && (enable == 1)) { pwm_disabled[dev->pin] = 0; }
+    if (enable == 1) {
+        pwm_disabled[dev->pin] = 0;
+    } else {
+        pwm_disabled[dev->pin] = 1;
+        // Set duty cycle to 0 before disable pwm.
+        // Edison PWM output stuck at high if disabled during ON period
+        mraa_pwm_pulsewidth_us(dev, 0);
```
alext-mkrs: This means the user's duty setting will be lost after they run `enable(0)` or `write(0)`, which will break existing code relying on the setting preservation. I'd say we should avoid that at all costs.

src/x86/intel_edison_fab_c.c
```diff
@@ -373,8 +374,17 @@ mraa_intel_edison_aio_init_post(mraa_aio_context dev)
...
+        // Sleep 2 period to allow change take effect
+        usleep(dev->period >> 9);
```
alext-mkrs: This means we are adding an additional delay, which again can break or alter the behavior of the existing code.

Are you at all sure the reason for this "stuck at 100%" is that the PWM is being disabled during the duty cycle? How did you come to that conclusion? My measurements (see comment in #585) do not confirm that as the failure rate should have been increasing with the duty increase, but it hasn't.

Finally - interesting way to calculate the two periods' time - not providing the exact 2x value AFAICS, why not simply multiply? Shifting is typically faster, but in this context it's unlikely to make any difference, plus that imprecision... I wonder what's the idea behind this.

belugon: I agree the delay is not desirable, do you recommend closing this PR?

From my view your test result mostly agree with this observation. 100% failure for 100%duty case, and 0% failure for 10% duty case. The lack of smooth transition could be caused by deterministic timing between operations.

For the last part, I agree it's unlikely going to make any difference, both in terms of speed as well as the timing (from my test, I stopped seeing any failure once delay reaches ~2 * period), so I went with my first instinct. I'll change this to the verbose way.

src/x86/intel_edison_fab_c.c
```diff
@@ -373,8 +374,17 @@ mraa_intel_edison_aio_init_post(mraa_aio_context dev)
 mraa_result_t
 mraa_intel_edison_pwm_enable_pre(mraa_pwm_context dev, int enable) {
     // PWM 0% duty workaround: update state array
-    // if someone first ran write(0) and then enable(1).
-    if ((pwm_disabled[dev->pin] == 1) && (enable == 1)) { pwm_disabled[dev->pin] = 0; }
+    if (enable == 1) {
+        pwm_disabled[dev->pin] = 0;
+    } else {
+        pwm_disabled[dev->pin] = 1;
```
belugon: I misunderstood the intent of this array. I'm changing it back.

alext-mkrs:
Okay, thanks for corrections, so now the two main points I have are:
1. Whether this is a proper solution in terms of "making it realiably switch off" - and your tests suggest that yes. Let me do some testing as well to see how it's going to behave on my setup, I'll  report back here when done.
1. The solution you propose adds a delay to switch off and changes the duty cycle, both can make the existing code to break or lead to other unexpected [for the user] effects. And that's something I'm worried about the most, as we want to be able to work around the HW problems to make user experience better, but at the same time we want to keep the mraa's API and behaviour stable, so that users trust it. @arfoll, what would be your take at that?

alext-mkrs:
On the test side (#1 above) - in my tests it also worked fine all the time. Nice finding about the "disable during active" being the reason, now that indeed looks like it.

alext-mkrs:
So @belugon, would you be willing to rework it to include the duty cycle restoration?

alext-mkrs:
@belugon, so anything you want to do on this one? If not, I may probably find the time this weekend.

alext-mkrs:
Okay, so I've added that duty cycle save/restore functionality I was talking about to this same PRs (nice recent functionality from GitHub). @arfoll, I've kept it as a separate commit for now as I'm not sure how you would want to merge those in terms of authorship. Could you please review and merge or let me know if you want anything else to be done here?

I've tested it both using my scripts I shared in #585 and manually, to make sure the duty is saved and restored - worked fine.

alext-mkrs:
@arfoll, do you think we should do anything else on this one?

arfoll:
Nope this is perfect. Merged!

alext-mkrs:
Thanks!

SathornModel:
Is there any update on this issue?

alext-mkrs:
As you can see it's merged a while ago.

SathornModel:
Thank you.  I haven't realized that the mraa on the default installation is not updated.  I run the following and it is up to 1.6.0, then the problem solve.

echo "src mraa-upm http://iotdk.intel.com/repos/3.5/intelgalactic/opkg/i586" > /etc/opkg/mraa-upm.conf
opkg install mraa

alext-mkrs:
Yep, that's the way to go, glad it works for you now.

PR #590, opened 2016-10-16
Add UART functionality to mock

alext-mkrs:
This is the last piece I was planning for the mock functionality, rounding up the feature set to include pretty much everything mraa can do (exceptions are PWM and GPIO ISRs, which I believe irrelevant for mock).

As usual, while doing that I've spotted a couple of unrelated small bugs/inconsistencies, which I piggyback as separate commits.

Of particular interest may be the one, which corrects the ` setNonBlocking()` function name - it was inconsistently spelled as `SetNonBlocking()`. Formally that's an API change, so that might bring some harm to existing code. Let's see if you think we can get it in right now or not. If not, then we could either leave both of spelling versions for some time or rework my mock functionality patch to use the incorrect version and only release the correction the next time we feel it's okay to go.

### DIFF
--- a/api/mraa/uart.hpp
+++ b/api/mraa/uart.hpp
@@ -263,7 +263,7 @@ class Uart
      * @return Result of operation
      */
     Result
-    SetNonBlocking(bool nonblock)
+    setNonBlocking(bool nonblock)
     {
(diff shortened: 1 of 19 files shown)


arfoll:
Merged! Thanks alot

PR #597, opened 2016-11-04
update Joule pin labels and add some new ones

jewalt:
Update pin 1 to match Joule labeling

Change Pins 100-104 to match Joule labeling

Add GPIO pins on 81-84

Add basic debian bundling


rwaldron:
> Change Pins 100-104 to match Joule labeling

Slightly tangential: which labels are you referring to?

arfoll:
@tingleby has been working on Debian packages for our ppa. But ur pin naming changes look interesting. Can you split the commit so the debian stuff is separate?

jewalt:
Sure, I can split the commit.  Pins 100-104 are labeled LED1, LED2, LED3, LED4 which is how the tuchuck board is silkscreened, but the signal names on the schematic (silkscreen directives?_ and the hardware design doc call those pins GPIO_0, GPIO_1, GPIO_2 and GPIO_3.  I thought it would be best to call the pins by the correct name.

rwaldron:
Sorry to bug you about this, but which is the "tuchuck board"? I have the Joule and expansion board, neither of which has the word "tuchuck" on it (keep in mind that I'm a third party that does not have access to internal intel manufacturing naming). 

My board has them labeled GP100, GP101, GP102, GP103

jewalt:
I think "Tuchuck" is internal name of the expansion board... At least that is what I have taken it to mean based on some of the documentation I have found.  I think those pins are label GPIO0, GPIO1, GPIO2 and GPIO3.  It just looks like GP100,GP101,GP102 and GP103.

rwaldron:
> GPIO0, GPIO1, GPIO2 and GPIO3

Ha! Indeed they are. 


So, is there now an expansion board that has them labelled LED1, LED2, LED3, LED4?

jewalt:
> So, is there now an expansion board that has them labelled LED1, LED2, LED3, LED4?

Maybe there was at some point, but I am not aware of one, thus my change.  The schematic for the breakout board has them labeled as in your picture so I'm not quite sure where the LED1, etc. came from.

rwaldron:
Ok, cool—thanks for explaining.

arfoll:
@jewalt do you know what, that's very true, and I don't think anyone has noticed ;-) Internally we've been calling them 100-103 for *ages*. I don't want to change it now, that'll lead to way more confusion than I need. I'll add a note in the documentation

@rwaldron 'expansion board' == 'tuchuck'. I think 'expansion board' is prone to confusion so I don't use it.

PR #611, opened 2016-11-27
Added Bubblegum 96 Support

ProfFan:
This is my portings for the 96boards Bubblegum-96 board support. Tested on latest linaro build with i2c-dev.

### DIFF
--- a/src/arm/96boards.c
+++ b/src/arm/96boards.c
@@ -25,67 +25,74 @@
  * WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
  */
 
-#include <stdlib.h>
+#include <mraa/common.h>
 #include <stdarg.h>
+#include <stdlib.h>
 #include <string.h>
 #include <sys/mman.h>
-#include <mraa/common.h>
 
-#include "common.h"
 #include "arm/96boards.h"
+#include "common.h"
 
-#define DT_BASE "/sys/firmware/devicetree/base"
+#define DT_BASE "/proc/device-tree"
 
 #define PLATFORM_NAME_DB410C "DB410C"
 #define PLATFORM_NAME_HIKEY "HIKEY"
+#define PLATFORM_NAME_BBGUM "BBGUM"
 
 int db410c_ls_gpio_pins[MRAA_96BOARDS_LS_GPIO_COUNT] = {
-  36, 12, 13, 69, 115, 4, 24, 25, 35, 34, 28, 33,
+    36, 12, 13, 69, 115, 4, 24, 25, 35, 34, 28, 33,
 };
 
-const char* db410c_serialdev[MRAA_96BOARDS_LS_UART_COUNT] = { "/dev/ttyMSM0", "/dev/ttyMSM1"};
+const char* db410c_serialdev[MRAA_96BOARDS_LS_UART_COUNT] = { "/dev/ttyMSM0", "/dev/ttyMSM1" };
 
 int hikey_ls_gpio_pins[MRAA_96BOARDS_LS_GPIO_COUNT] = {
- 488, 489, 490, 491, 492, 415, 463, 495, 426, 433, 427, 434,
+    488, 489, 490, 491, 492, 415, 463, 495, 426, 433, 427, 434,
 };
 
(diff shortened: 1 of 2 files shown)


src/arm/96boards.c
```diff
@@ -34,23 +34,31 @@
...
-#define DT_BASE "/sys/firmware/devicetree/base"
+#define DT_BASE "/proc/device-tree"
 
 #define PLATFORM_NAME_DB410C "DB410C"
 #define PLATFORM_NAME_HIKEY "HIKEY"
+#define PLATFORM_NAME_BBGUM "BBGUM"
 
 int db410c_ls_gpio_pins[MRAA_96BOARDS_LS_GPIO_COUNT] = {
-  36, 12, 13, 69, 115, 4, 24, 25, 35, 34, 28, 33,
+   36, 12, 13, 69, 115, 4, 24, 25, 35, 34, 28, 33,
```
alext-mkrs: Whitespace error at the beginning of the string

src/arm/96boards.c
```diff
@@ -34,23 +34,31 @@
...
 };
 
 const char* db410c_serialdev[MRAA_96BOARDS_LS_UART_COUNT] = { "/dev/ttyMSM0", "/dev/ttyMSM1"};
 
 int hikey_ls_gpio_pins[MRAA_96BOARDS_LS_GPIO_COUNT] = {
- 488, 489, 490, 491, 492, 415, 463, 495, 426, 433, 427, 434,
+   488, 489, 490, 491, 492, 415, 463, 495, 426, 433, 427, 434,
```
alext-mkrs: Whitespace error at the beginning of the string

src/arm/96boards.c
```diff
@@ -34,23 +34,31 @@
...
 };
 
+
```
alext-mkrs: Surplus stying edit

src/arm/96boards.c
```diff
@@ -34,23 +34,31 @@
 #include "common.h"
 #include "arm/96boards.h"
 
-#define DT_BASE "/sys/firmware/devicetree/base"
+#define DT_BASE "/proc/device-tree"
```
alext-mkrs: [Disclaimer: I don't have any of these boards] Are you sure this won't break other boards supported by this file? Is that the case that this path is now `/proc/device-tree` on all those boards and their respective distros/versions?

src/arm/96boards.c
```diff
@@ -97,7 +105,12 @@ mraa_board_t* mraa_96boards()
             ls_gpio_pins = hikey_ls_gpio_pins;
             b->uart_dev[0].device_path = hikey_serialdev[0];
             b->uart_dev[1].device_path = hikey_serialdev[1];
-	}
+	     } else if (mraa_file_contains(DT_BASE "/model", "s900")) {
```
alext-mkrs: This should be on a separate line (hint: run `clang-format` on your changes using config file from the root of the repo)

src/arm/96boards.c
ProfFan: I have not tried these boards, but this path is incorrect itself. As stated by the Linux ABI, (https://www.kernel.org/doc/Documentation/ABI/testing/sysfs-firmware-ofw),
>  Userspace must not use the `/sys/firmware/devicetree/base` path directly, but instead should follow `/proc/device-tree` symlink. It is possible that the absolute path will change in the future, but the symlink is the stable ABI.

I believe using this path is a common misconception in the community.

src/arm/96boards.c
ProfFan: Will fix ASAP.

ProfFan:
I have not tried these boards, but this path is incorrect itself. As stated by the Linux ABI, (https://www.kernel.org/doc/Documentation/ABI/testing/sysfs-firmware-ofw),
>  Userspace must not use the `/sys/firmware/devicetree/base` path directly, but instead should follow `/proc/device-tree` symlink. It is possible that the absolute path will change in the future, but the symlink is the stable ABI.

I believe using this path is a common misconception in the community.

All source files have been `clang-format`ed.

Added `Signed-off-by`.

BTW, I have a bunch other 96boards types on the way. Will add them when they arrive 😄

arfoll:
This is great, thanks alot! Merged.

Issue #502, opened 2016-05-01
After upgrading to mraa 1.0 node fails to locate libmraa.so.0


ramonmata:
Hello,
I just upgraded my Intel Edison to the 3.0 Yocto release, so then upgraded all upgradeable opkg packages to the latest.

Then when trying the example for sensor BISS0001, node complains about a missing libmraa.so.0 file

**libmraa.so.0: cannot open shared object file: No such file or directory**

```
Intel XDK - IoT App Daemon v0.0.38 - commands: run, list, debug, status
Upload on board is not from current project - uploading complete project
No NPM modules found.
transferring /var/folders/m9/6wtq5xqx2rqgs46rq03_g_mh0000gn/T/xdk-9708nad0c8g.tar from project to board
x README.md
x icon.png
x main.js
x package.json
x xdk/project-info.json
Upload Complete

ERROR: module.js:356
ERROR:   Module._extensions[extension](this, filename);
   ERROR:                             ^
ERROR: Error: libmraa.so.0: cannot open shared object file: No such file or directory
```

I was able to solve this problem by creating a symbolic link with that name in `/usr/lib` pointing to `libmraa.so.1` but perhaps this is a issue that needs to be solved

best regards!

arfoll:
We just need a upm release to fix. @propanu?

hellofunk:
I am having this identical problem. I'm a bit of a linux newbie, can you advise on the actual steps you took at the edison ssh prompt to resolve this?

ramonmata:
Try this:

```
cd /usr/lib
ln -s libmraa.so.1 libmraa.so.0
```

regards

arfoll:
Thanks @ramonmata. Note that there may be some issues due to the versioning/API change in some sensors though I expect the effects to be very limited.

alext-mkrs:
it's not very clear from the error message, so @arfoll, you think that's mraa Node binding problem or UPM problem, to make sure? In view of your ping for @propanu, I'd assume the latter, but maybe mraa releases are now not done by you?

arfoll:
Mraa releases are done by me, this API change was done on purpose since mraa's API changed slightly with 1.0.0. The next version of UPM will fix this since they will link to libmraa.so.1.

Propanu:
Yes, a UPM update will fix this issue, we're working on it and 0.7.0 will be out soon.

alext-mkrs:
Understood, thanks.

deadprogram:
[...]

PR #631, opened 2017-01-18
UartOW C++ interface fixes

jontrulson:
uart_ow.hpp: Correct a logic bug, and 2  questionable uses of c_str()
    
    The command() function had a logic error that caused a command to be
    broadcast to all devices on the DS OW bus when a proper ID was
    specified.  This should only be done when NO ID is specified.
    
    This fixes UPM issue #502.


arfoll:
Merged, thanks alot!

PR #632, opened 2017-01-19
joule: enable ISH_GPIOs and GPIO_22 exposed on J12 header

Propanu:

### DIFF
--- a/src/x86/intel_gt_tuchuck.c
+++ b/src/x86/intel_gt_tuchuck.c
@@ -151,9 +151,9 @@ mraa_gt_tuchuck_board()
     b->pins[pos].capabilities = (mraa_pincapabilities_t){ 0, 0, 0, 0, 0, 0, 0, 0 };
     pos++;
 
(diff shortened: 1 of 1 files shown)


arfoll:
This is good but your strncpy limit character write to 8-1 yet you give strings that are 8+1. You need to change your naming to match the string length, I recommend just making them ISHIOn rather than ISHGPIOn?

Propanu:
Forgot about the terminator. I'm ok with ISHIOn. In fact I think ISH_IOn will be a bit more readable and still fit. Alternatively we can also bump the limit up to MRAA_PIN_NAME_SIZE. Let me know which option you prefer.

arfoll:
Let's just use ISH_IOn, increasing MRAA_PIN_NAME_SIZE will eat up more mem, if we can fit it in let's just do that :)

Propanu:
Should be all good now!

arfoll:
rebased as one and merged. Thanks!

PR #637, opened 2017-01-31
added close of file handle

machielbruinink:
Will testing multiple i2c devices running in multiple threads we noticed that closing and opening a device would leave behind an open file descriptor we added one line of code to ensure the file descriptor is closed properly when we close a i2c device.


arfoll:
This is a great fix but in order for us to be able to accept it you need to sign it off - just run `git commit --amend --sign-off` and `git push --force`. Thanks!

arfoll:
Merged, thanks alot for the fixup!

PR #643, opened 2017-02-07
fix compilation issues

spitfire88:
Rename remaining B* -> A*


arfoll:
Thanks for the quick fix. Can we add a travis.ci target for AT?

PR #644, opened 2017-02-08
Add support for UART hardware flow control CTS RTS

nicolalunghiemutex:
add support for hardware flow control pin in uart pin  definition
Then each platform can configure those pin or the uart library can enable or disable them based on uart mode selected

### DIFF
--- a/include/mraa_internal_types.h
+++ b/include/mraa_internal_types.h
@@ -340,6 +340,8 @@ typedef struct {
     unsigned int index; /**< ID as exposed in the system */
     int rx; /**< uart rx */
     int tx; /**< uart tx */
+    int cts; /**< uart cts */
+    int rts; /**< uart rts */
     char* device_path; /**< To store "/dev/ttyS1" for example */
     /*@}*/
 } mraa_uart_dev_t;


arfoll:
Not that I mind the change, but I'm curious to see what's next, as it stands, this doesn't seem particularly useful and would just add some compiler warnings about unused struct members?

nicolalunghiemutex:
I am developing support for a new platform and I need to know which pin are the CTS/RTS to configure it before the standard uart init function 
(using the _pre and _post functions), and unexport them if they are configured as GPIO
Later could be added support for CTS/RTS in libmraa?

arfoll:
Sure, and I'm happy to merge this with the new platform as like you say it's the right way to do it, however on it's own this commit brings no value.

nicolalunghiemutex:
ok

nicolalunghiemutex:
the problem is if you use hardware flow control the pin are not initialized

see file
 **https://github.com/nicolalunghiemutex/mraa/blob/master/src/uart/uart.c**

on function
```
mraa_uart_context
mraa_uart_init(int index)

...

 if (!plat->no_bus_mux) {
        int pos = plat->uart_dev[index].rx;
        if (pos >= 0) {
            if (plat->pins[pos].uart.mux_total > 0) {
                if (mraa_setup_mux_mapped(plat->pins[pos].uart) != MRAA_SUCCESS) {
                    syslog(LOG_ERR, "uart%i: init: failed to setup muxes for RX pin", index);
                    return NULL;
                }
            }
        }
```

here checks if the rx (and tx) pin are mapped as gpio but it doesn't check the CTS/RTS pins
so we need a way to know if there are CTS/RTS pin defined and if you want to use the 
hardware flow control module you need to configure those pins also
I don't know if the best place to put this is in the init function (but we need to know the mode before the init) or in the set_flowcontrol function

```
mraa_result_t
mraa_uart_set_flowcontrol(mraa_uart_context dev, mraa_boolean_t xonxoff, mraa_boolean_t rtscts)
{
    if (!dev) {
        syslog(LOG_ERR, "uart: set_flowcontrol: context is NULL");
        return MRAA_ERROR_INVALID_HANDLE;
    }

    if (IS_FUNC_DEFINED(dev, uart_set_flowcontrol_replace)) {
        return dev->advance_func->uart_set_flowcontrol_replace(dev, xonxoff, rtscts);
    }

    // hardware flow control
    int action = TCIOFF;
    if (xonxoff) {
        action = TCION;
    }
    if (tcflow(dev->fd, action)) {
        return MRAA_ERROR_FEATURE_NOT_SUPPORTED;
    }

    // rtscts
    struct termios termio;

    // get current modes
    if (tcgetattr(dev->fd, &termio)) {
        syslog(LOG_ERR, "uart%i: set_flowcontrol: tcgetattr() failed: %s", dev->index, strerror(errno));
         return MRAA_ERROR_INVALID_RESOURCE;
    }

    if (rtscts) {
        termio.c_cflag |= CRTSCTS;
    } else {
        termio.c_cflag &= ~CRTSCTS;
    }

    if (tcsetattr(dev->fd, TCSAFLUSH, &termio) < 0) {
        syslog(LOG_ERR, "uart%i: set_flowcontrol: tcsetattr() failed: %s", dev->index, strerror(errno));
        return MRAA_ERROR_FEATURE_NOT_SUPPORTED;
    }

    return MRAA_SUCCESS;
}
```

arfoll:
Totally agree, and if your code would modify those functions to check the RTS/CTS values I would merge it in an instant ;-) The problem is right now all this does is add two members to a struct that on their own do very little and generally I don't merge new features unless the whole feature works in one logical PR (which may be lots of small commits).

PR #648, opened 2017-02-14
Adding a static library target in CMakeLists.txt

anonymouse64:
Adding additional static library target for linking statically against libmraa.


anonymouse64:
Forgot to sign-off my commit... I will re-open with a signed off commit.

PR #649, opened 2017-02-14
Adding additional static library target that shows up as libmraa.a

anonymouse64:
Adding additional static library target for linking statically against libmraa.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -255,7 +255,9 @@ if (BUILDSWIG)
   endif ()
 endif ()
 
-add_library (mraa ${mraa_LIB_SRCS})
+add_library (mraa SHARED ${mraa_LIB_SRCS})
+add_library (mraa_static STATIC ${mraa_LIB_SRCS})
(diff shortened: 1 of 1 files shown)


arfoll:
So this forces cmake to build both rather than what the system vars/cmake toolchain file decides, so I'm not sure I like it. At the same time, it's handy. Would you object to having this under a cmake var like -DBUILDBOTH=ON (or if you have a better name) ? And having the default stay the same?

anonymouse64:
Yes, that would be sensible as most cases you probably just want the dynamic library, not the static one

arfoll:
So if all you want is to have just a static library you can just use ` -DBUILD_SHARED_LIBS=OFF`. That's why we moved from specifying DYNAMIC to not specifying anything and letting users choose. I've added a note in the docs with fc03ea4816264ca9f900b42dda2b369209cb8543.

In any case if you think having both is useful and guard it with an option flag I'm happy to merge that.

anonymouse64:
Ah I wasn't aware of this CMake option, this accomplishes the same task, I don't think that building both is necessary. Thanks

PR #171, opened 2015-03-14
make uninstall work

tripzero:
This enables 'make uninstall'

Signed-off-by: Kevron Rees <tripzero.kev@gmail.com>

PR #652, opened 2017-02-21
Added Uninstall command to remove installed files

nicolalunghiemutex:
also check if in the python directories are present .pyc files and removes them

Signed-off-by: Nicola Lunghi <nicola.lunghi@emutex.com>


nicolalunghiemutex:
Seen now that mostly the same file was added here

https://github.com/intel-iot-devkit/mraa/pull/171

and then removed?

if its useful I can re-commit it
Nick

PR #653, opened 2017-02-23
Update references to the Intel Joule

w4ilun:


alext-mkrs:
Please sign-off your commits and ideally squash the two, that merge one is just surplus.

w4ilun:
Apologies, PR updated

docs/joule.md
```diff
@@ -1,7 +1,7 @@
-Intel Joule {#grossetete}
+Intel Joule
```
alext-mkrs: I think this one needs `{#joule}`, just like the old version had `{#grossetete}`, as it's used as the anchor by the docs generator and it's a corresponding "endpoint" for `@ref joule` in `index.md`

w4ilun:
Changes added

tingleby:
Any reference in users code that makes reference to MRAA_INTEL_JOULE_EXPANSION would no longer work. This should be considered an API Break.

alext-mkrs:
LGTM with the corrections and @tingleby definitely has a point. Maybe we could define both old and new symbols for now, warn users in release note and remove the old one whenever we feel appropriate to break the API?

alext-mkrs:
Oh and just one more thing - there are currently two commits in the PR with identical names, @w4ilun you'd want to squash them to have just one.

arfoll:
Thanks for the PR why, I've thought about doing this for a while, although I think there's a few good reasons to not rename as I worry about the existing documentation links (they should be versioned but not everyone is good ;-)) as well as the enum value changing technically causing an API break.

w4ilun:
Thanks @arfoll , I submitted this PR because there were a few requests regarding consistent naming in our docs; the breaking API changes need not be included, we can just update the docs & links. Let me know what you guys think.

PR #667, opened 2017-03-10
Joule shield with ADC detection

Propanu:
Description pretty much sums it up. If you have a Grove or DFRobot Shield for Joule and are running the latest kernel, this will detect it and load the appropriate IIO driver, configure it, and finally, add the AIO pins to the pinmap.

A couple of other minor changes:
 - reordered I2C buses
 - added UART buses + device paths

### DIFF
--- a/src/x86/intel_joule_expansion.c
+++ b/src/x86/intel_joule_expansion.c
@@ -185,6 +185,15 @@ mraa_joule_expansion_board()
     b->spi_bus[4].bus_id = 32765;
     b->spi_bus[4].slave_s = 2;
 
+    b->uart_dev_count = 2;
+    b->def_uart_dev = 0;
+    b->uart_dev[0].device_path = "/dev/ttyS0";
+    b->uart_dev[0].rx = 68;
+    b->uart_dev[0].tx = 7;    
+    b->uart_dev[1].device_path = "/dev/ttyS1";
+    b->uart_dev[1].rx = 24;
+    b->uart_dev[1].tx = 22;
+
     // Aio shield detection
     // Only if i2c_designware.0 was successfully found
     if (i2c_aio_bus != -1) {
--- a/src/x86/intel_joule_expansion.c
+++ b/src/x86/intel_joule_expansion.c
@@ -245,7 +245,8 @@ mraa_joule_expansion_board()
 
                         // max sample rate and 6.144 V reference
                         // additional aio functions can be added to make up for loss of precision at 3V3, 5V, etc
-                        for (int i = 0; i < b->aio_count; i++) {
+                        int i;
+                        for (i = 0; i < b->aio_count; i++) {
                             snprintf(command, 128, "echo 3300 >/sys/bus/iio/devices/iio:device0/in_voltage%d_sampling_frequency", i);
                             system(command);
                             snprintf(command, 128, "echo 3 >/sys/bus/iio/devices/iio:device0/in_voltage%d_scale", i);
@@ -279,7 +280,8 @@ mraa_joule_expansion_board()
 
                         // max sample rate and 6.144 V reference
                         // additional aio functions can be added to make up for loss of precision at 3V3, 5V, etc
-                        for (int i = 0; i < b->aio_count; i++) {
+                        int i;
+                        for (i = 0; i < b->aio_count; i++) {
                             snprintf(command, 128, "echo 860 >/sys/bus/iio/devices/iio:device%d/in_voltage%d_sampling_frequency", i / 4, i % 4);
                             system(command);
                             snprintf(command, 128, "echo 3 >/sys/bus/iio/devices/iio:device%d/in_voltage%d_scale", i / 4, i % 4);
--- a/src/x86/intel_joule_expansion.c
+++ b/src/x86/intel_joule_expansion.c
@@ -109,22 +109,6 @@ mraa_joule_expansion_board()
         b->i2c_bus_count++;
     }
 
-    i2c_bus_num = mraa_find_i2c_bus_pci("0000:00", "0000:00:17.1", "i2c_designware.5");
-    if (i2c_bus_num != -1) {
-        b->i2c_bus[b->i2c_bus_count].bus_id = i2c_bus_num;
-        b->i2c_bus[b->i2c_bus_count].sda = 15;
-        b->i2c_bus[b->i2c_bus_count].scl = 17;
-        b->i2c_bus_count++;
-    }
-
-    i2c_bus_num = mraa_find_i2c_bus_pci("0000:00", "0000:00:17.2", "i2c_designware.6");
-    if (i2c_bus_num != -1) {
-        b->i2c_bus[b->i2c_bus_count].bus_id = i2c_bus_num;
-        b->i2c_bus[b->i2c_bus_count].sda = 19;
-        b->i2c_bus[b->i2c_bus_count].scl = 21;
-        b->i2c_bus_count++;
-    }
-
     i2c_bus_num = mraa_find_i2c_bus_pci("0000:00", "0000:00:16.1", "i2c_designware.1");
     if (i2c_bus_num != -1) {
         b->i2c_bus[b->i2c_bus_count].bus_id = i2c_bus_num;
@@ -138,6 +122,22 @@ mraa_joule_expansion_board()
         b->i2c_bus[b->i2c_bus_count].bus_id = i2c_bus_num;
         b->i2c_bus[b->i2c_bus_count].sda = 75;
         b->i2c_bus[b->i2c_bus_count].scl = 77;
+        b->i2c_bus_count++;
+    }
+
+    i2c_bus_num = mraa_find_i2c_bus_pci("0000:00", "0000:00:17.1", "i2c_designware.5");
+    if (i2c_bus_num != -1) {
+        b->i2c_bus[b->i2c_bus_count].bus_id = i2c_bus_num;
+        b->i2c_bus[b->i2c_bus_count].sda = 15;
+        b->i2c_bus[b->i2c_bus_count].scl = 17;
+        b->i2c_bus_count++;
+    }
+
+    i2c_bus_num = mraa_find_i2c_bus_pci("0000:00", "0000:00:17.2", "i2c_designware.6");
+    if (i2c_bus_num != -1) {
+        b->i2c_bus[b->i2c_bus_count].bus_id = i2c_bus_num;
+        b->i2c_bus[b->i2c_bus_count].sda = 19;
+        b->i2c_bus[b->i2c_bus_count].scl = 21;
         b->i2c_bus_count++;
     }
 
--- a/src/x86/intel_joule_expansion.c
+++ b/src/x86/intel_joule_expansion.c
@@ -29,7 +29,7 @@
 #include <sys/types.h>
 #include <sys/stat.h>
 #include <sys/ioctl.h>
-#include <linux/i2c-dev.h>
+#include <mraa/i2c.h>
 
 // Debug
 #include <syslog.h>
@@ -39,20 +39,6 @@
 #include "x86/intel_joule_expansion.h"
 
 #define PLATFORM_NAME "INTEL JOULE EXPANSION"
-
-typedef union i2c_smbus_data_union {
-    uint8_t byte;        ///< data byte
-    unsigned short word; ///< data short word
-    uint8_t block[I2C_SMBUS_BLOCK_MAX + 2];
-    ///< block[0] is used for length and one more for PEC
-} i2c_smbus_data_t;
-
-typedef struct i2c_smbus_ioctl_data_struct {
-    uint8_t read_write;     ///< operation direction
-    uint8_t command;        ///< ioctl command
-    int size;               ///< data size
-    i2c_smbus_data_t* data; ///< data
-} i2c_smbus_ioctl_data_t;
 
 typedef enum {NO_SHIELD, DFROBOT, GROVE} shield_t;
 
@@ -99,7 +85,6 @@ mraa_joule_expansion_board()
 
     b->i2c_bus_count = 0;
 
-    //int i2c_bus_num = -1;
     int i2c_bus_num = mraa_find_i2c_bus_pci("0000:00", "0000:00:16.0", "i2c_designware.0");
     int i2c_aio_bus = i2c_bus_num;
     if (i2c_bus_num != -1) {
@@ -189,7 +174,7 @@ mraa_joule_expansion_board()
     b->def_uart_dev = 0;
     b->uart_dev[0].device_path = "/dev/ttyS0";
     b->uart_dev[0].rx = 68;
-    b->uart_dev[0].tx = 7;    
+    b->uart_dev[0].tx = 7;
     b->uart_dev[1].device_path = "/dev/ttyS1";
     b->uart_dev[1].rx = 24;
     b->uart_dev[1].tx = 22;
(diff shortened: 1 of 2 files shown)


src/x86/intel_joule_expansion.c
```diff
@@ -145,6 +181,115 @@ mraa_joule_expansion_board()
...
+
+        char i2c_fh[64];
+        snprintf(i2c_fh, 64, "/dev/i2c-%d", i2c_aio_bus);
+
+        sm.read_write = I2C_SMBUS_READ;
+        sm.command = 0x01;
+        sm.size = I2C_SMBUS_WORD_DATA; // LSB first
+        sm.data = &d;
+
+        if ((fp = open(i2c_fh, O_RDWR)) < 1)
```
arfoll: You must be joking right? We have the mraa i2c raw API just for this kind of stuff... If it doesnt do what you need, we need to fix it!

src/x86/intel_joule_expansion.c
```diff
@@ -24,48 +25,83 @@
...
     b->pwm_default_period = 5000;
     b->pwm_max_period = 218453;
     b->pwm_min_period = 1;
 
     b->i2c_bus_count = 0;
 
-    int i2c_bus_num = -1;
-    i2c_bus_num = mraa_find_i2c_bus_pci("0000:00", "0000:00:16.0", "i2c_designware.0");
+    //int i2c_bus_num = -1;
+    int i2c_bus_num = mraa_find_i2c_bus_pci("0000:00", "0000:00:16.0", "i2c_designware.0");
```
arfoll: We've already indexed them all?

src/x86/intel_joule_expansion.c
```diff
@@ -24,48 +25,83 @@
...
+    uint8_t block[I2C_SMBUS_BLOCK_MAX + 2];
+    ///< block[0] is used for length and one more for PEC
+} i2c_smbus_data_t;
+
+typedef struct i2c_smbus_ioctl_data_struct {
+    uint8_t read_write;     ///< operation direction
+    uint8_t command;        ///< ioctl command
+    int size;               ///< data size
+    i2c_smbus_data_t* data; ///< data
+} i2c_smbus_ioctl_data_t;
```
arfoll: urgh no, why do we need this stuff?

src/x86/intel_joule_expansion.c
```diff
@@ -24,48 +25,83 @@
...
+
+typedef enum {NO_SHIELD, DFROBOT, GROVE} shield_t;
+
+static mraa_result_t
+mraa_joule_expansion_board_get_valid_fp(mraa_aio_context dev)
+{
+    char file_path[64] = "";
+
+    // 4 channels per ADC
+    snprintf(file_path, 64, "/sys/bus/iio/devices/iio:device%d/in_voltage%d_raw", dev->channel / 4, dev->channel % 4);
```
arfoll: Isn't this what the aio module does? And if not let's fix it?

src/x86/intel_joule_expansion.c
```diff
@@ -24,48 +25,83 @@
...
+} i2c_smbus_data_t;
+
+typedef struct i2c_smbus_ioctl_data_struct {
+    uint8_t read_write;     ///< operation direction
+    uint8_t command;        ///< ioctl command
+    int size;               ///< data size
+    i2c_smbus_data_t* data; ///< data
+} i2c_smbus_ioctl_data_t;
+
+typedef enum {NO_SHIELD, DFROBOT, GROVE} shield_t;
```
arfoll: platform_version is meant do exactly this

src/x86/intel_joule_expansion.c
```diff
@@ -145,6 +181,115 @@ mraa_joule_expansion_board()
...
+            // no DFRobot shield, try Grove
+            if (ioctl(fp, I2C_SLAVE_FORCE, 0x48) < 0)
+                syslog(LOG_ERR, "Failed to set I2C address: %s", strerror(errno));
+            if (ioctl(fp, I2C_SMBUS, &sm) < 0)
+                syslog(LOG_NOTICE, "No device at I2C address 0x48");
+                // no shield
+            else {
+                // load ads1015 at 0x48
+                syslog(LOG_NOTICE, "Loading ti-ads1015 module for Grove Shield");
+                if (system("modprobe ti-ads1015") < 0) {
```
arfoll: I'm really not sure magic modprobing is the way.  However doing system( ) calls like this are definately the wrong way, linux/module.h has things to do this.

src/x86/intel_joule_expansion.c
```diff
@@ -145,6 +181,115 @@ mraa_joule_expansion_board()
...
+                        b->adc_supported = 10;
+                        pincount += b->aio_count;
+                        shield = GROVE;
+
+                        // max sample rate and 6.144 V reference
+                        // additional aio functions can be added to make up for loss of precision at 3V3, 5V, etc
+                        for (int i = 0; i < b->aio_count; i++) {
+                            snprintf(command, 128, "echo 3300 >/sys/bus/iio/devices/iio:device0/in_voltage%d_sampling_frequency", i);
+                            system(command);
+                            snprintf(command, 128, "echo 3 >/sys/bus/iio/devices/iio:device0/in_voltage%d_scale", i);
```
arfoll: Again random system() calls should not be something we do. No reason we can't do this with the Aio class?

arfoll (changes requested):

src/x86/intel_joule_expansion.c
Propanu: We probably don't, I'll see if it can be removed with the proper includes added.

src/x86/intel_joule_expansion.c
Propanu: Not sure I follow here. This is only used internally to keep track of how many ADC channels we have.

src/x86/intel_joule_expansion.c
Propanu: The aio module doesn't handle multiple devices. In fact, it assumes iio:device0 is the ADC. Proper way would be to actually query all the IIO devices and figure out which ones are the ADCs. That's a pretty substantial "fix".

src/x86/intel_joule_expansion.c
Propanu: I'm a bit lost here, are you thinking about i2c_bus_count?

src/x86/intel_joule_expansion.c
Propanu: I2C raw would do exactly what's needed, however there isn't a single board definition that relies on the API headers. If that's something you encourage, I see no problem using the raw functions instead. It would also get rid of the smbus data structs since they would be pulled from MRAA's i2c.h...

src/x86/intel_joule_expansion.c
Propanu: This is the tricky part. Modprobe wouldn't be needed if the ads1015 wasn't double aliased (hwmon/ads1015 vs. iio/ti-ads1015). I tend to disagree with using linux/module.h as we'll just end up rewriting what insmod and modprobe already do. The system call with a hardcoded string I'm not worried about, the rest of them on the other hand I would happily replace with a i2c_new_device() + MRAA's IIO functions if all needed functionality is there.

arfoll:
It's absolutely what our internal i2-dev header is for and pretty sure it
contains all you need see
https://github.com/intel-iot-devkit/mraa/blob/master/include/linux/i2c-dev.h

On 14 Mar 2017 04:13, "Mihai Tudor Panu" <notifications@github.com> wrote:

*@Propanu* commented on this pull request.
------------------------------

In src/x86/intel_joule_expansion.c
<https://github.com/intel-iot-devkit/mraa/pull/667#discussion_r105819846>:

>
 #include "common.h"
 #include "x86/intel_joule_expansion.h"

 #define PLATFORM_NAME "INTEL JOULE EXPANSION"

+typedef union i2c_smbus_data_union {
+    uint8_t byte;        ///< data byte
+    unsigned short word; ///< data short word
+    uint8_t block[I2C_SMBUS_BLOCK_MAX + 2];
+    ///< block[0] is used for length and one more for PEC
+} i2c_smbus_data_t;
+
+typedef struct i2c_smbus_ioctl_data_struct {
+    uint8_t read_write;     ///< operation direction
+    uint8_t command;        ///< ioctl command
+    int size;               ///< data size
+    i2c_smbus_data_t* data; ///< data
+} i2c_smbus_ioctl_data_t;

We probably don't, I'll see if it can be removed with the proper includes
added.

—
You are receiving this because you commented.

Reply to this email directly, view it on GitHub
<https://github.com/intel-iot-devkit/mraa/pull/667#discussion_r105819846>,
or mute the thread

arfoll:
Sure but we can use platform version for this and then expose extra info to
users

arfoll:
Agreed it's not great, but we can just make that iio device string platform
mappable and then use it. That shouldn't be a big fix and remove a load of
code from this...

arfoll:
Uhm yes not sure what i saw there either my bad :)

arfoll:
We have lots of platform defs that use the gpio raw API i see no difference
with that and the i2c API. Please use it

alext-mkrs:
I also wonder is sub-platform functionality would be a better fit for this, rather then packing the shield-related code into base expansion board file? Or maybe a copy of this file, with this additional logic, could be treated as a seperate platform...

Propanu:
Cleaned up and resolved most of the suggestions. Additional change:
* Enabled use of mraa_i2c_init_raw() during platform detection.

I think the `mraa_joule_expansion_board_get_valid_fp()` replacement is valid for now, but can be improved by expanding on the new `aio_dev` struct that was recently added. Will spend some time on that, since the ADC class needs more flexibility.

On the other hand, loading and exporting drivers would be better of in the iio class as a new feature, but I doubt I'll find time for that.

@arfoll & @alext-mkrs what are your thoughts?

src/x86/intel_joule_expansion.c
```diff
@@ -149,6 +170,114 @@ mraa_joule_expansion_board()
...
+                if (system("modprobe ti-ads1015") < 0) {
+                    syslog(LOG_NOTICE, "Failed, are you running the latest Joule kernel?");
+                } else {
+                    char command[128];
+                    snprintf(command, 128, "echo ads1115 0x48 >/sys/bus/i2c/devices/i2c-%d/new_device", i2c_aio_bus);
+                    if (system(command) < 0)
+                        syslog(LOG_ERR, "Failed to add ads1115 device");
+                    else {
+                        snprintf(command, 128, "echo ads1115 0x49 >/sys/bus/i2c/devices/i2c-%d/new_device", i2c_aio_bus);
+                        if (system(command) < 0)
```
alext-mkrs: Even single-line `if` should have braces per mraa's accepted code style

src/x86/intel_joule_expansion.c
```diff
@@ -149,6 +170,114 @@ mraa_joule_expansion_board()
...
+                }
+            } else {
+                // load ads1115 at 0x48 and 0x49
+                syslog(LOG_NOTICE, "Loading ti-ads1015 module for DFRobot Shield");
+                if (system("modprobe ti-ads1015") < 0) {
+                    syslog(LOG_NOTICE, "Failed, are you running the latest Joule kernel?");
+                } else {
+                    char command[128];
+                    snprintf(command, 128, "echo ads1115 0x48 >/sys/bus/i2c/devices/i2c-%d/new_device", i2c_aio_bus);
+                    if (system(command) < 0)
```
alext-mkrs: Even single-line `if` should have braces per mraa's accepted code style

src/x86/intel_joule_expansion.c
```diff
@@ -149,6 +170,114 @@ mraa_joule_expansion_board()
...
+    b->uart_dev[1].tx = 22;
+
+    // Aio shield detection
+    // Only if i2c_designware.0 was successfully found
+    if (i2c_aio_bus != -1) {
+
+        syslog(LOG_NOTICE, "Attempting shield autodetection on I2C bus %d...", i2c_aio_bus);
+        mraa_i2c_context i2c = mraa_i2c_init_raw(i2c_aio_bus);
+
+        if (i2c == NULL)
```
alext-mkrs: Even single-line `if` should have braces per mraa's accepted code style

src/x86/intel_joule_expansion.c
```diff
@@ -149,6 +170,114 @@ mraa_joule_expansion_board()
...
+            syslog(LOG_ERR, "Failed to open I2C bus: %d", i2c_aio_bus);
+        else {
+            if (mraa_i2c_address(i2c, 0x49) != MRAA_SUCCESS)
```
alext-mkrs: Even single-line `if` should have braces per mraa's accepted code style

src/x86/intel_joule_expansion.c
```diff
@@ -149,6 +170,114 @@ mraa_joule_expansion_board()
...
+                syslog(LOG_ERR, "Failed to set I2C address: %d", 0x49);
+            if (mraa_i2c_read_word_data(i2c, 0x01) < 0) {
+                syslog(LOG_NOTICE, "No device at I2C address 0x49");
+
+                // no DFRobot shield, try Grove
+                if (mraa_i2c_address(i2c, 0x48) != MRAA_SUCCESS)
```
alext-mkrs: Even single-line `if` should have braces per mraa's accepted code style

src/x86/intel_joule_expansion.c
```diff
@@ -149,6 +170,114 @@ mraa_joule_expansion_board()
...
+                    syslog(LOG_ERR, "Failed to set I2C address: %d", 0x48);
+                if (mraa_i2c_read_word_data(i2c, 0x01) < 0)
```
alext-mkrs: Even single-line `if` should have braces per mraa's accepted code style

src/x86/intel_joule_expansion.c
```diff
@@ -149,6 +170,114 @@ mraa_joule_expansion_board()
...
+                    syslog(LOG_NOTICE, "No device at I2C address 0x48");
+                    // no shield
+
+                else {
+                    // load ads1015 at 0x48
+                    syslog(LOG_NOTICE, "Loading ti-ads1015 module for Grove Shield");
+                    if (system("modprobe ti-ads1015") < 0) {
```
alext-mkrs: Loading modules just doesn't seem the right thing to do for the library like mraa - but I think that was already mentioned by @arfoll in the previous review :)

Maybe we'd rather tell the user to do that (in the documentation), as they surely know which shield is attached?

alext-mkrs (changes requested):
There's a bunch of pure style-related comments (single-line `if`s), but generally it's fine if we assume that we are ok with doing `modprobe` and friends from the library. I'm still not convinced this is the Right Way, I'd rather let user do this by clearly documenting this. That way they can set all those proper corrections and whatever else settings are necessary for ADCs to work.

src/x86/intel_joule_expansion.c
arfoll: yes I agree, we shouldn't module load and definately no system call. Please replace this with a check for the module being loaded and working and document you need it.

src/x86/intel_joule_expansion.c
```diff
@@ -149,6 +170,114 @@ mraa_joule_expansion_board()
...
+                        syslog(LOG_NOTICE, "Failed, are you running the latest Joule kernel?");
+                    } else {
+                        char command[128];
+                        snprintf(command, 128, "echo ads1015 0x48 >/sys/bus/i2c/devices/i2c-%d/new_device", i2c_aio_bus);
```
arfoll: This is lazyness, we should not be creating calls to then run on the shell. Please use standard methods from C to do this

src/x86/intel_joule_expansion.c
```diff
@@ -149,6 +170,114 @@ mraa_joule_expansion_board()
...
+                            b->adc_supported = 10;
+                            pincount += b->aio_count;
+                            shield = GROVE;
+
+                            // max sample rate and 6.144 V reference
+                            // additional aio functions can be added to make up for loss of precision at 3V3, 5V, etc
+                            int i;
+                            for (i = 0; i < b->aio_count; i++) {
+                                snprintf(command, 128, "echo 3300 >/sys/bus/iio/devices/iio:device0/in_voltage%d_sampling_frequency", i);
+                                system(command);
```
arfoll: We really really shouldn't do this this way, this is horrible. It also assumes there is no other iio:device0, you can use the mraa IIO api to query by device name and do this much more nicely.

arfoll (changes requested):
The `system()` need to be removed, assumptions that iio:device0 is the one are also bad and should be changed.

src/x86/intel_joule_expansion.c
Propanu: So do we want to remove the SPI driver loading from the RPi board file then too?

arfoll: Now you got me, I'd forgotten about those :) I do actually feel like it's not something a library should do. We should probably remove it from all platforms and document it but I'd like to hear more opinions so, @alext-mkrs @tingleby @pylbert @Hbrinj @w4ilun @jontrulson - I trust you all, do you think in the name of user experience we should do this or not?

jontrulson: Hmm.

Well, as a first observation, **IF** you want to do this kind of thing I'd recommend 2 things:
1. Use fork/exec rather than system().  Since all you want to do is call an executable, there is no point in invoking an intermediate shell.  It's more complicated, particularly in detecting an error (you'll need a sig handler), but will be faster and safer.  Since it's not done very often though, this is probably overkill.
2. Regardless of which method you use to spawn the process, **ALWAYS** use a complete path to the executable you are spawning -- in this case **_/sbin/modprobe_**.  This avoids a whole host of "Hacking 101" possibilities via PATH or other environment manipulation.

As for whether it should be done at all in this case - I am definitely a fan of "It should Just Work®".  But in this case the functionality depends on external kernel modules being loaded (or even available), so I think I am with @arfoll and @alext-mkrs here:  This should be done at system startup or manually by the user (documentation should make this clear).  MRAA should expect everything it needs from the kernel/OS to be there and be ready.

In my view, it would make sense to just have all of these potential drivers be loaded at boot (via modules.conf or whatever), _assuming_ of course that MRAA can reliably detect at runtime whether these devices are in fact available (hardware present/driver loaded), and use them if they are.

alext-mkrs: I'm for just documenting the necessary steps, providing helper scripts, whatever else can make user's life easier - but not doing that from the library itself. And yeah, let's remove that from Pi3 code as well, that must've slipped through everyone as I also don't recall seeing that (I'd have objected) :)

Propanu: So I guess I'm fine then changing this to use iio instead, then modprobe + friends can go into documentation. Should we start an issue and move this thread there then, requiring additional work for the RPi? I'd hate to see all these good comments go to waste when I push this branch again 😄

arfoll: #770 submitted.

@Propanu I think it's pretty clear cut, let's document module usage and remove these. I also want to remove all the system() calls generally from this PR. To avoid loosing the history, maybe just create a new PR? That would likely be best :)

PR #668, opened 2017-03-10
Fix issue with gpio-339 mapping

arunlee77:
The mapping for gpio-339 went away with commit

gpio-339 is available as LED102 or ISH_IO2


arfoll:
docs need an update? Please add a prefix like other commits, in this case "joule: " or "intel_joule_expansion.c: " is probably suitable since it's only that file.

Propanu:
@arunlee77 thanks for fixing this!

arfoll:
Merged!

PR #671, opened 2017-03-13
uart_ow: prevent possible infinite loop in _ow_read_byte()

jontrulson:
With a broken device, or a broken interface circuit, it is possible
for _ow_read_byte() to spin forever.  This patch arranges for it to
timeout after 5 seconds if it cannot read a valid byte from the UART.

### DIFF
--- a/src/uart_ow/uart_ow.c
+++ b/src/uart_ow/uart_ow.c
@@ -35,7 +35,7 @@
 #include <termios.h>
 #include <fcntl.h>
 #include <errno.h>
-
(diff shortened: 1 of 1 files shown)


alext-mkrs:
Sorry I can't add a line comment doing this on the phone - but the new code doesn't have braces around one-line if statements, which is what mraa common style is.

jontrulson:
Fixed, thanks!

arfoll:
LGTM. Merged!

PR #672, opened 2017-03-13
joule: Fix issue with ISH UART name

arunlee77:
MRAA is using gpio 484, 483, 485 and 486 as ISH UART1.
But J13 expansion connector doesn't expose ISH UART1,
instead it exposes ISH UART0 as per dev kit hardware guide.

This patch fixes this descrpency and renames the UART and
also enables the GPIO usage.


arfoll:
LGTM. merged.

PR #673, opened 2017-03-14
mraa.c: Fix strsep crash

skuusela:
Variable 'dup' doesn't get freed correctly as the pointer to it changes
after 'strsep' function is used. Also 'token' variable doesn't need to
be freed as freeing the original 'dup' already frees the same memory.

Signed-off-by: Simo Kuusela <simo.kuusela@intel.com>


skuusela:
Got help for this from @ipuustin.

arfoll:
Thanks alot, this is great. merged!

PR #675, opened 2017-03-14
Periphmraa

stefan-andritoiu:


stefan-andritoiu:
The error were generated by some implicit casts that the compiler did not like. The question remains as to why this is a warning when building mraa, but an error when trying to build periphmraa.

arfoll:
Merged. Probably should merge periphmraa now.

PR #678, opened 2017-03-20
intel_joule: Update doc with correct PIN behavior

arunlee77:
GPIO and I2C functions of some PINs are not possible
with default BIOS configuration. Current documentation
wrongly shows that the PINs can work as both GPIO and I2C.

This patch fixes this issue and also updated pin conf for
I2C 1 and 2 to disable GPIO functionality.


arunlee77:
Tested with v1.5.1 in ref-os-iot with BIOS 193.

arfoll:
Travis still being silly, merged!

PR #682, opened 2017-03-23
Update platform_checks.py

nicolalunghiemutex:
Resolve the error on platform different than edison


nicolalunghiemutex:
Sorry I forgot
now should be ok

alext-mkrs:
Why do we need the PlatformChecks class, if it'sjust "pass"?

nicolalunghiemutex:
To remember to put something in it :-)
The problem is that the current platform check code fails on my board -> they are only for edison I think?
and we have a test bench for it and every time complains about the failed test

arfoll:
Merged

PR #684, opened 2017-03-23
de10-nano: minor readme update and added to types.hpp

Propanu:


arfoll:
Merged. I'm really not sure ALTERA_SOCFPGA is a good platform name, SOC FPGA sounds a bit too generic to me...

Propanu:
I agree on that! I'll submit another patch since it shouldn't be used anywhere yet.

PR #686, opened 2017-03-24
java: bypass version check for packages that don't have version specified

stefan-andritoiu:
Signed-off-by: Stefan Andritoiu <stefan.andritoiu@gmail.com>


arfoll:
Merged, thanks!

PR #690, opened 2017-03-29
mraa.c: Allow building MOCKPLAT

pylbert:
Changed order of #if statement for using mock platform.

If DETECTED_ARCH is "i586", "x86_64", or "i386", then X86PLAT is set,
which appears to be mutually exclusive with using the mock platform.

This commit gives priority to MOCKPLAT.

### DIFF
--- a/src/mraa.c
+++ b/src/mraa.c
@@ -132,15 +132,15 @@ imraa_init()
 
     // Not an else because if the env var didn't load what we wanted maybe we can still load something
     if (platform_type == MRAA_NULL_PLATFORM) {
-#if defined(X86PLAT)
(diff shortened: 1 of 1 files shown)


alext-mkrs:
LGTM, makes sense.

pylbert:
I see that the recommended way to enable the MOCK platform is to use -DBUILDARCH=MOCK.  Closing this.

alext-mkrs:
Generally, yes, that's why I implemented the detection that way. But that isn't any hard rule, I did that to make sure people know what they do and they really intend to build for mock. So if you see any additional value in rearranging detections like you proposed in this PR, why not.

PR #691, opened 2017-03-30
Deinit and leaks

pylbert:


arfoll:
Another gcc attribute... not sure super useful but looks funky so lets do it :)

PR #694, opened 2017-04-05
Create IndexLookup functions for GPIO, I2C, SPI, PWM.

vintummala:
Create IndexLookup functions for GPIO, I2C, SPI, PWM.
These lookups provide the MRAA index with the Pin/Bus name as input


spitfire88 (commented):
Fix compiler warnings. Declaration of 'int i' should must be outside the for loop.

In include/mraa_internal_types.h, conditionally compile "char *name" for PERIPHERALMAN.

spitfire88:
Additional comments from bjbeare to move the name field under each bus_t structures.

PR #696, opened 2017-04-05
java: bypass version check for packages that don't have version speci…

vintummala:
…fied

Signed-off-by: Vineela Tummalapalli <vineela.tummalapalli@intel.com>


stefan-andritoiu:
1. 
The method mraa.class.getPackage().getSpecificationVersion() will return NULL if version is not set. It will only return 0.0 if the version is explicitly set to 0.0. I don't know if it should bypass if the version is 0.0. Maybe someone wanted a specific version to be 0.0.

2.
Also, *==* and *!=* tests for reference equality (whether they are the same object), not if they have equal value. 
The statement *getSpecificationVersion()  != "0.0" * will always be true because the two objects will always be different. 
You mean to use * !getSpecificationVersion().equals("0.0")* . But DO NOT use .equals() if getSpecificationVersion() is null.

3.
Also, this commit will clash with the existing commit in master branch
d2bbfe376da661498c1a1224309ab4fec85910b9

stefan-andritoiu:
1.
The method mraa.class.getPackage().getSpecificationVersion() will return NULL if version is not set. It will only return 0.0 if the version is explicitly set to 0.0. I don't know if it should bypass if the version is 0.0. Maybe someone wanted a specific version to be 0.0.

2.
Also, == and != tests for reference equality (whether they are the same object), not if they have equal value.
The statement  getSpecificationVersion() != "0.0"  will always be true because the two objects will always be different.
You mean to use  !getSpecificationVersion().equals("0.0") . But DO NOT use .equals() if getSpecificationVersion() is null.

3.
Also, this commit will clash with the existing commit in master branch
d2bbfe376da661498c1a1224309ab4fec85910b9

PR #702, opened 2017-04-07
periphmraa: Create IndexLookup functions for GPIO, I2C, SPI, PWM.

vintummala:
These lookups provide the MRAA index with the Pin/Bus name as input.

Signed-off-by: Vineela Tummalapalli <vineela.tummalapalli@intel.com>

### DIFF
--- a/api/mraa/common.h
+++ b/api/mraa/common.h
@@ -234,6 +234,42 @@ unsigned int mraa_get_platform_pin_count(uint8_t platform_offset);
 char* mraa_get_pin_name(int pin);
 
 /**
+* Get pin number, board must be initialised.
+*
+* @param pin_name: GPIO Pin Name. Eg: IO0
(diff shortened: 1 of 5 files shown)


vintummala:
@arfoll 
For one of your comments "2.	The C++ based functions should throw exceptions", I understand i should throw exceptions in the common.hpp file.
But i do not see any of the apis in that file throwing exceptions.
I have come across an api in common.hpp file which does have negative returns on failure, but does not throw exception as below
setPriority(const int priority)
{
    return mraa_set_priority(priority);
}

Please let me know if the comment I addressed in common.hpp is the expectation?

arfoll:
Agreed, more functions like these in C++ should throw exceptions, the ones in the I/O modules are much better. This does need improving. Merged with some minor alterations.

PR #709, opened 2017-04-21
docs: Added markdown to docs folder for 96Boards

sdrobertw:
Initial run at 96Boards development platform markdown has been added.
Started with introduction, board support, interface notes and pin mapping.

Signed-off-by: Robert Wolff <robert.wolff@linaro.org>


arfoll:
This is great, can we remove the merge and have that as one commit? I don't think it needs so many :)

PR #654, opened 2017-02-23
Refactored JS examples for clarity, consistent style, and Node.JS fea…

nodebotanist:
…tures

Signed-off-by: Kassandra Perch <the@nodebotani.st>

PR #710, opened 2017-04-23
JS examples: refactor for clarity and consistent style

alext-mkrs:
Here goes the corrected version of PR #654, with my review comments addressed and some more fixes. Turned out on e.g. Edison additional "use strict" is required as Node.js version there is rather old and needs that. This shouldn't harm anyway and on newer version turns into no-op (it's enabled by default).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,20 +2,12 @@ language: cpp
 sudo: required
 dist: trusty
 env:
-  - NODE010=true
-  - NODE012=true
   - NODE4=true
   - NODE5=true
-  - JSONPLAT=true NODE010=true
-  - JSONPLAT=true NODE012=true
   - JSONPLAT=true NODE4=true
   - JSONPLAT=true NODE5=true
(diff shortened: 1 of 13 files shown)


alext-mkrs:
Oh and BTW, these `let` and friends don't work on Node 0.10 and 0.12, so I wonder if we still want to support those versions at all (maybe that's a good time to remove them from `travis.yml`). What do you think, @arfoll? I think even Galileo has the same as Edison (4.4.x) these days.

arfoll:
@alext-mkrs yes I'm fine with our examples breaking node 0.10 and 0.12 compat. I think this is acceptable.

alext-mkrs:
Thanks! Let me then bundle a `travis.yml` update as well, as a separate commit, to get rid of 0.10 and 0.12.

alext-mkrs:
Ok, Travis config corrected as well. If there are no objections (@arfoll, @whbruce, @propanu), I'll merge this one on Monday.

PR #711, opened 2017-04-25
periphmraa: Fix PWM issues

spitfire88:
Update pin capabilities for each supported board.
Do it statically since there is no way to retrieve information
about the relationship between the pins gpio and alternate
functionality from peripheral manager.

### DIFF
--- a/include/mraa_adv_func.h
+++ b/include/mraa_adv_func.h
@@ -90,7 +90,6 @@ typedef struct {
     mraa_result_t (*pwm_init_pre) (int pin);
     mraa_result_t (*pwm_init_post) (mraa_pwm_context pwm);
     mraa_result_t (*pwm_period_replace) (mraa_pwm_context dev, int period);
-    mraa_result_t (*pwm_duty_cycle_replace) (mraa_pwm_context dev, int duty_cycle);
     float (*pwm_read_replace) (mraa_pwm_context dev);
     mraa_result_t (*pwm_write_replace) (mraa_pwm_context dev, float duty);
     mraa_result_t (*pwm_write_pre) (mraa_pwm_context dev, float percentage);
--- a/include/mraa_adv_func.h
+++ b/include/mraa_adv_func.h
@@ -86,6 +86,7 @@ typedef struct {
 
     mraa_pwm_context (*pwm_init_replace) (int pin);
     mraa_pwm_context (*pwm_init_internal_replace) (void* func_table, int pin);
+    mraa_result_t (*pwm_init_raw_replace) (mraa_pwm_context dev, int pin);
     mraa_result_t (*pwm_init_pre) (int pin);
     mraa_result_t (*pwm_init_post) (mraa_pwm_context pwm);
     mraa_result_t (*pwm_period_replace) (mraa_pwm_context dev, int period);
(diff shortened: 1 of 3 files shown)


spitfire88:
Agree. I will continue to look at how this could be done in Peripheral Manager.

spitfire88:
MRAA does the static bit in src/x86/intel_*.c files.

bjbeare:
yes, but this is p-mraa. :)

vintummala:
@spitfire88 
I see update_pincapabilities_edison() defined two times.
I think the other function name should be update_pincapabilities_joule() which you called in update_pincapabilitie().
Did'nt it throw compilation error?

spitfire88:
@vintummala It didn't. Do you know how to set PERIPHERALMAN during compilation in cmake? 
Guessing the DETECTED_ARCH variable needs to be set somehow.

vintummala:
@spitfire88 
https://docs.google.com/document/d/1kKwAoM7KwfreXWWETk1neArf_zdpZ8GkcRk_lWSpW0Y/edit#heading=h.nrhuzc3r159j
this document should work fine to compile MRAA with androidthings.
Bruce edited this and is more clear now :)

bjbeare:
Yes... I had to work around some issues that need to be fixed PLUS the cmake.txt files need to set reasonable defaults.
Next up... getting the .aar created

spitfire88:
After discussion with Bruce, updated the PR.

bjbeare:
Looks much better, thank you.

Can I get a test version of the updated .aar so that I can verify this fix>

vintummala:
Now that we are commenting out the code verification of capabilities for PERIPHMRAA, the corresponding code which uses capabilities needs to be under if !defined(PERIPHMRAA).
Looks like mraa_pin_mode_test() in mraa.c is doing checks with capabilities and it is throwing compilation error as below. please add this under "if !defined(PERIPHMRAA)"

/home/brillo/PeripheralManager-folder/mraa-mainline/mraa/src/mraa.c:623:41: error: no member named 'capabilities' in 'mraa_pininfo_t'
            if (current_plat->pins[pin].capabilities.valid == 1)
                ~~~~~~~~~~~~~~~~~~~~~~~ ^
/home/brillo/PeripheralManager-folder/mraa-mainline/mraa/src/mraa.c:627:41: error: no member named 'capabilities' in 'mraa_pininfo_t'
            if (current_plat->pins[pin].capabilities.gpio == 1)
                ~~~~~~~~~~~~~~~~~~~~~~~ ^
/home/brillo/PeripheralManager-folder/mraa-mainline/mraa/src/mraa.c:631:41: error: no member named 'capabilities' in 'mraa_pininfo_t'
            if (current_plat->pins[pin].capabilities.pwm == 1)

arfoll:
Just a tought - instead of #ifdef everywhere, would it not be easier to just have peripheralmanager platform memset the capabilities to '1'?

spitfire88:
@arfoll Yes that was the initial commit to address this issue. Bruce and I later agreed that the capabilities feature which is taken care of by Peripheral Manager should be entirely skipped for this platform. Although that leads to the ugly #ifdef's. And plenty of them.

In my opinion the short term solution would be to memset the capabilities to 1 for the pins and the long term solution would be to get that information from peripheral manager (by adding a new API for board layout information). That way we keep MRAA clean.

@bjbeare any thoughts?

bjbeare:
I think that's a reasonable and more elegant solution.

bjbeare:
+2 on merging this.
There's still a frequency issue... but at least the pin selection is working. We'll handle the frequency problem with a separate pull request.

PR #722, opened 2017-05-08
peripheralman: default to board's I2C frequency

spitfire88:
PIO does not provide an API to set frequency. Default to board's
I2C frequency.

### DIFF
--- a/src/peripheralman/peripheralman.c
+++ b/src/peripheralman/peripheralman.c
@@ -41,6 +41,30 @@ int uart_busses_count = 0;
 char **pwm_devices = NULL;
 int pwm_dev_count = 0;
 
+typedef enum {
+    EDISON_ARDUINO = 0,
+    EDISON_SPARKFUN = 1,
+    JOULE_TUCHUCK = 2,
+    INVALID = 3
+} Board;
+
+static unsigned int
+get_board_variant(void)
+{
+    if (gpios == NULL) {
+        return INVALID;
+    }
+
+    if (!strncmp("IO0", gpios[0], 3)) {
+        return EDISON_ARDUINO;
+    } else if (!strncmp("GP12", gpios[0], 4)) {
+        return EDISON_SPARKFUN;
+    } else if (!strncmp("J6_1", gpios[0], 4)) {
+        return JOULE_TUCHUCK;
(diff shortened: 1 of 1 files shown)


src/peripheralman/peripheralman.c
```diff
@@ -412,7 +412,15 @@ mraa_pman_i2c_stop_replace(mraa_i2c_context dev)
 static mraa_result_t
 mraa_pman_i2c_set_frequency_replace(mraa_i2c_context dev, mraa_i2c_mode_t mode)
 {
-    return MRAA_ERROR_FEATURE_NOT_SUPPORTED;
+    switch (mode) {
```
bjbeare: Need a comment here that we default to the board's frequency but that when PeripheralManager supports the call... we'll fill in the action routines.

pylbert:
If a board's default frequency is unknown and it's not possible to set it via the PIO, this should return an error code (MRAA_ERROR_FEATURE_NOT_SUPPORTED).

If the I2C default frequency is guaranteed, I think it would be OK to return success only for the matching case.

spitfire88:
While your comments are applicable to Edison how do I handle this on Joule? 
Sensors looking to change frequency mode just fail on Joule and I am not familiar of any way to do that.

spitfire88:
@Propanu I have made the requested edits. Please review. Thanks.

Propanu:
Quick note I2C_FAST is default on Joule too.

spitfire88:
Oh sweet! I will make that change. Is there a sysfs node that prints this information?

arfoll:
I still don't get this. We can't change the i2c frequency therefore returning MRAA_ERROR_FEATURE_NOT_SUPPORTED *is* the correct thing to do. I'm not merging code that simply confuses the API. This function is meant to change the speed. If it can't then it can't.

You can check in sysfs for edison here: /sys/devices/pci0000:00/0000:00:08.0/i2c_dw_sysnode/mode. As far as I know that's a modified kernel driver so not generic.

In any case 99% of boards are in i2c_fast mode in default kernel configurations. The only exception I know to this is galileo gen1 which is in 100khz because of the i2c I/O multiplexer.

PR #726, opened 2017-05-11
peripheralman: fix return values

spitfire88:
Be consistent with the return values from MRAA and propagate the
errors received from PIO in case of failure. PIO returns 0 on success
and negative errno on failure. Handle APIs that must return length of
read/written data if PIO returns success.

### DIFF
--- a/src/peripheralman/peripheralman.c
+++ b/src/peripheralman/peripheralman.c
@@ -102,7 +102,7 @@ mraa_pman_pwm_enable_replace(mraa_pwm_context dev, int enable)
 static float
 mraa_pman_pwm_read_replace(mraa_pwm_context dev)
 {
-    return -MRAA_ERROR_FEATURE_NOT_SUPPORTED;
+    return -1;
 }
 
 static mraa_result_t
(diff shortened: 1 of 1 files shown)


spitfire88:
I wanted the return value to be less generic as -1, in case the handle was NULL. But I do agree and Noel was in agreement as well, that it hurts the eye. I could revert back to -1 if that works? A positive MRAA_ERROR_INVALID_HANDLE return, which would be 7, could be misinterpreted as read/writes being completed. Which is not the case here so that's out of the question. Let me know.

arfoll:
sadly -1 is the best we can do, we've never had negative error values. Not really the best system I'll grant you, but the way to do it better in C for these cases is to set errno which I always thought was kinda ugly, so I prefer to give nice logs in syslog.

If someone is serious about doing this you'd have to do the whole lot it can't be just a few random APIs :). If you're really keen on this I propose you make a good proposal and let everyone comment.

spitfire88:
@arfoll I will talk internally with my team if this can be added to the to do list. In the case of Android development better error reporting will help debug issues faster.

For this patch though I have reverted back to -1. Please review.

arfoll:
So generally this now returns -1 or if there is a return code from a PIO function it returns that? What will PIO return generally be in case of an error?

spitfire88:
PIO logs a service specific error if the exception is in Java followed by an errno. While in native it just returns errno. Here is a reference https://android.googlesource.com/kernel/lk/+/upstream-master/include/errno.h.

bjbeare:
@arfoll ... Are you satisfied with this current change set?
Please let us know if an additional change is being requested... else let's merge!

arfoll:
Sorry about delay - merged.

PR #729, opened 2017-05-13
Some stability fixes to prevent potential buffer overflows

alext-mkrs:
Though unlikely in these specific pieces of code, it's an industry BKM to use `*n*` versions of `strcpy()` and `sprintf()` to avoid buffer overflows. We also use `*n*` ones across the board, so let's align these several outliers too.

### DIFF
--- a/examples/mraa-i2c.c
+++ b/examples/mraa-i2c.c
@@ -249,9 +249,11 @@ run_interactive_mode()
         if (strcmp(command, "q") == 0)
             return;
         char* str = strtok(command, " ");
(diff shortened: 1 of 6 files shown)


alext-mkrs:
I'll merge this one later today if there are no objections, @arfoll, @Propanu, @whbruce

alext-mkrs:
Thanks @whbruce!

Committed with 0654183b9a83428a97909724aa49a1e29535bd62 and 9545a2e3207d387733992f1c47462b310ef548e3

PR #730, opened 2017-05-16
mraa: clean up and bugfix by-name device lookup functions; add uart lookup function

TechNexion:
PR 707 introduced new and useful functions to look-up devices using a text string instead of index.
There are three patches in this PR related to these functions.

Patch 1. Adapts the style in some of the functions introduced by PR707 to match the rest of the codebase (like return type on a separate line). No functional changes.
Patch 2. Fixes potential null pointer errors on platforms where ever bus/device/pin is not named, or when a null pointer is passed as argument (user error).
Patch 3. Introduces a similar function to look up UARTs by name. Somehow UART lookup was omitted in the earlier patches.

### DIFF
--- a/api/mraa/common.h
+++ b/api/mraa/common.h
@@ -266,6 +266,14 @@ int mraa_spi_lookup(const char* spi_name);
 int mraa_pwm_lookup(const char* pwm_name);
 
 /**
+* Get UART index by name, board must be initialised.
+*
+* @param uart_name: Name of UART. Eg:UART1
+* @return int of MRAA index for UART, or -1 if not found.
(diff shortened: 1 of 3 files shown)


arfoll:
Thanks for this - it's great. Merged (with very minor doc styling fixes added since you did a styling fix already :)).

Note that the strcmp functions need to be changed for strncmp, but @alext-mkrs is doing this already in a bunch of other areas and all the lookup functions fail in this reguard - @alext-mkrs can you handle those?

alext-mkrs:
Yep, no problem.

Issue #731, opened 2017-05-16
strcmp usage instead of strncmp in _lookup functions


arfoll:
Assigning to @alext-mkrs since you're looking at a bunch of these - if you can't do it I will happily take a look.

alext-mkrs:
Sure, will do, no problem.

PR #733, opened 2017-05-17
Replace strcmp() with strncmp()

alext-mkrs:
This is for #731. Arguably in some of the cases we don't really need `*n*` version (string literals) or just trade one potential segmentation fault (from unbound `strcmp()`) to another (from unbound `strlen()` - `strnlen()` would probably be an answer here, but I'm not sure we want to go down that path).

But even if the above holds, it still makes sense from the code unification standpoint - we use `strncmp()` everywhere else.

Side note - I've noticed in most of the cases we use `strncmp()`, we are actually doing a substring comparisons (i.e. without a null-terminator taken into account). While this apparently works ok - as no one complained so far and nothing broke - I could go over and replace that with exact match checks. Let me know what you think about that.

### DIFF
--- a/examples/mraa-i2c.c
+++ b/examples/mraa-i2c.c
@@ -166,16 +166,16 @@ int
 process_command(int argc, char** argv)
 {
     int status = 0;
-    if (strcmp(argv[1], "help") == 0) {
+    if (strncmp(argv[1], "help", strlen("help") + 1) == 0) {
         print_help();
         return 0;
-    } else if (strcmp(argv[1], "version") == 0) {
+    } else if (strncmp(argv[1], "version", strlen("version") + 1) == 0) {
         print_version();
         return 0;
-    } else if (strcmp(argv[1], "list") == 0) {
+    } else if (strncmp(argv[1], "list", strlen("list") + 1) == 0) {
         print_busses();
         return 0;
-    } else if (strcmp(argv[1], "detect") == 0) {
+    } else if (strncmp(argv[1], "detect", strlen("detect") + 1) == 0) {
         if (argc == 3) {
             int bus = strtol(argv[2], NULL, 0);
             i2c_detect_devices(bus);
@@ -184,10 +184,11 @@ process_command(int argc, char** argv)
             print_command_error();
             return 1;
         }
-    } else if ((strcmp(argv[1], "get") == 0) || (strcmp(argv[1], "getrpt") == 0)) {
+    } else if ((strncmp(argv[1], "get", strlen("get") + 1) == 0) ||
+               (strncmp(argv[1], "getrpt", strlen("getrpt") + 1) == 0)) {
         if (argc == 5) {
             int interation = 0;
-            mraa_boolean_t should_repeat = strcmp(argv[1], "getrpt") == 0;
+            mraa_boolean_t should_repeat = strncmp(argv[1], "getrpt", strlen("getrpt") + 1) == 0;
             int bus = strtol(argv[2], NULL, 0);
             uint8_t device_address = strtol(argv[3], NULL, 0);
             uint8_t register_address = strtol(argv[4], NULL, 0);
@@ -211,7 +212,7 @@ process_command(int argc, char** argv)
             status = 1;
         }
         return status;
-    } else if ((strcmp(argv[1], "set") == 0)) {
+    } else if ((strncmp(argv[1], "set", strlen("set") + 1) == 0)) {
         if (argc == 6) {
             int bus = strtol(argv[2], NULL, 0);
             uint8_t device_address = strtol(argv[3], NULL, 0);
@@ -246,7 +247,7 @@ run_interactive_mode()
         fprintf(stdout, "Command: ");
         fgets(command, 80, stdin);
         command[strlen(command) - 1] = 0;
-        if (strcmp(command, "q") == 0)
+        if (strncmp(command, "q", strlen("q") + 1) == 0)
             return;
         char* str = strtok(command, " ");
         int len = 0;
--- a/src/mraa.c
+++ b/src/mraa.c
@@ -833,7 +833,8 @@ mraa_gpio_lookup(const char* pin_name)
     }
 
     for (i = 0; i < plat->gpio_count; i++) {
-         if (plat->pins[i].name != NULL && strcmp(pin_name, plat->pins[i].name) == 0) {
+         if (plat->pins[i].name != NULL &&
+             strncmp(pin_name, plat->pins[i].name, strlen(plat->pins[i].name) + 1) == 0) {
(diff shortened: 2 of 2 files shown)


TechNexion:
Uh. When would `strncmp(a, b, strlen(b)+1)` avoid a problem that `strcmp(a, b)` would have? Can you give an example?
Hint: never. ;-)

Sorry to have to point out the obvious: `strcmp` will compare at most `MIN(strlen(a)+1, strlen(b)+1)` characters, and your `strncmp` at most `strlen(b)+1`
The "problem" with `strcmp` occur when **both** input strings are not (properly) terminated. And using `strlen` will not help you there.

A probably better way would be to have a `#define` that limits the maximum size of names for devices, and using that in `strncmp`. While cludgey, using any upper limit is still better than unbounded.

alext-mkrs:
Thanks for your comments. I know all of that and considered all points you have highlighted, when preparing the PR - thus my comment to it and stressing the "code unification" above everything else 

Additionally, for sizeof() vs strlen() specifically:
1) Compiler optimization replaces strlen() with a const anyway, when possible;
2) Sizeof will not properly work on pointers and may confuse relatively inexperienced developers, so this is for clarity and readability.

arfoll:
Good shout on the fact we dont actually check terminating char. That could bite us as gpio-1 would match gpio-11 etc... Probably worth fixing at some point.

TechNexion:
Wonder how Linus Torvalds would describe this ... ;-)

arfoll:
If you provide a better fix then you're welcome to submit it ;-) In the meantime if Linus is available for a code review, we'd obviously love that input.

TechNexion:
Changing things like:
`if (strcmp(argv[1], "help") == 0) {`
to
`if (strncmp(argv[1], "help", strlen("help") + 1) == 0) {`
is where I imagined Linus using some ... colourful ... langauge about it.

For constant strings like here, it does not make any sense at all. It does not make the code any better, any more efficient, any easier to read or write. Rather the opposite in all possible ways.
The only way I would touch that line would be to make it
`if (!strcmp(argv[1], "help")) {`
but that would going in the opposite direction. But hey, that's me, and I don't really feel it is my place to object.

alext-mkrs:
> Good shout on the fact we dont actually check terminating char. That could bite us as gpio-1 would match gpio-11 etc... Probably worth fixing at some point.

Submitted #736 to track this.

TechNexion:
Actually, the new version with `strncmp` is _strictly worse_ than the old with `strcmp` from a program robustness point-of-view as well.
`strncmp(a, b, strlen(b))` will go ballistic if one of the strings (here `b`) is badly terminated. `strcmp(a, b)` needs both to be badly terminated.
So for instance if `a` is `"\0"` (or any string shorter than the allocation for `b`) and `b` is unterminated (longer than `a`), `strcmp` will not crash while the `strncmp` version will. There is no case where `strcmp` will crash but `strncmp` will not.

alext-mkrs:
You make this statement based on what exactly? Looking at [`str{n}cmp()` sources in Linux](http://elixir.free-electrons.com/linux/latest/source/lib/string.c#L316), I see they use a null-terminator from the first string as the comparison loop exit marker, in both functions. And in `strncmp()` there's that `n` limit on top, which limits the number of cycles - and which in our code is based on the second string, so to me it looks like in our case we actually are a bit more protected, by having a limit based on both strings.

I haven't analyzed the arch-specific implementations, my assumption being that they're algorithmically equivalent.

TechNexion:
@alext-mkrs : Ok let's go back to school :-)
(Discussion about the point you raised after the proof)
**"Theorem":**
There are cases of a and b where `strncmp(a, b, strlen(b))` can fail in when `strcmp(a, b)`  cannot.
There are no cases where `strcmp(a, b)` can fail but `strncmp(a, b, strlen(b))` cannot.
**Proof:**
 Let OK denote no problem, and NG mean that there is chance for invalid memory access. Also let "terminated" mean that there is at least one null byte within the allocated block of a string.

Compare strcmp(a, b) vs strncmp(a, b, strlen(b)) and divide it into six cases based on the termination statuses of a and b as follows.

_i)_ a terminated, b terminated  => strcmp OK, strncmp OK
_ii)_ a unterminated, b terminated, b shorter (or equal) than a's allocation => strcmp OK, strncmp OK
_iii)_ a unterminated, b terminated, b longer than a's allocation => strcmp NG, strncmp NG
_iv)_ a terminated, b unterminated, a shorter (or equal) than b's allocation => strcmp OK, strncmp NG
_v)_ a terminated, b unterminated, a longer than b's allocation => strcmp NG, strncmp NG
_vi)_ a unterminated, b unterminated => strcmp NG, strncmp NG

In case _iv)_ strncmp can fail, but strcmp cannot. This proves the first statement. There are no cases above where strcmp fails but strncmp cannot. This proves the second statement.
_QED_

Regarding the argument about which null byte implementations use internally. 

It does not matter whether str{n}cmp function looks at the null byte of the first, second or both arguments. The comparison stops at the first null byte, regardless of which string it occurs in. If one string has a null char at position _k_ and the other not does not have a null char at position _k_, the strings differ at positon _k_ and the functions return. If both have a null byte, checking any one of them is enough.
With the strncmp way, the problem is moved to strlen. It can *always* fail for unterminated strings.

My thoughts of using strncmp instad of strcmp only changes to the worse in pretty much every way I can imagine:
- Code is longer
- Code is harder to read
- Code is harder to write
- Code is more inefficient
- Code is uglier (this is subjective)
- Code is less general
- Code is harder to maintain

So what is the benefit? How did this ... unusual ... idea even come by without anyone reacting?

While I don't feel this is my code to maintain, and there are strong emotions to use strncmp ...  go ahead. Use the gun. Aim at your foor. But sane, it is not. :-)

arfoll:
@TechNexion If your argument is that strcmp -> strncmp is these cases has little value I agree. And actually I've learnt quite abit about strcmp so that's been interesting :)

However I'm really not sure this PR it merits the - whilst passionate - long debate above. If you feel really strongly about it, fix all the strncmp that are not required and add a quick bit of doc that explains why strcmp is the appropriate choice so the next guy with static analysis doesn't take the easy path. I'll happily merge that PR :)

alext-mkrs:
@technexion, thanks for analysis, that formally proves my description of this PR, where I'm saying that we're trading one potential segfault for the other ;-) and that th main goal of the PR is code unification.

Now getting back to why we need this whole thing at all - to protect us from unterminated strings - the solution is really just to replace the strlen(string_of_unknown_length) with some hard cap a'la MAX_SIZE we use elsewhere.

Oh and yes, the whole topic is a bit moot as all of us know and understand what's going on, it's just the choice we made based on factors beyond strictly technical area.

PR #735, opened 2017-05-18
UART speed related additions/fixes

TechNexion:
This PR involves three patches, nothing major.

1. A function that converts a `speed_t` to integer baudrate (code using this function is not-yet commited).
2. `uint2speed()` function would not always return a `speed_t`, but the `int` constant 0. Changed that to the `speed_t` equivalent. This patch has probably no functional changes.
3. A function that returns the number of available UARTs 

There are two discussion points regarding patch 3.

1. Is there a reason for lack of similar accessor functions for other pins/buses (like gpio, spi, pwm ...)? Maybe there is a design goal to keep the interfaces small and not cluttered with too many functions? Or just nobody got around writing them? In the latter case, I have similar patches ready to be pushed for accessing other board items.
2. My 3rd patch also violates what appears to be a style standard in the framework. Instead of just declaring the function, I also provide a prototype for it. Maybe there is a reason for the lack of prototypes for functions of zero arity? (Esthetical? Not realizing the difference declaration and prototype?). The later C standards (C99 and onwards) strongly discourage the use of declarations the way they are used here (as far as I remember...).

So before I push more patches, it would be good if there is a hint about which way to go.

### DIFF
--- a/src/uart/uart.c
+++ b/src/uart/uart.c
@@ -110,7 +110,7 @@ uint2speed(unsigned int speed)
 #endif
         default:
             // if we are here, then an unsupported baudrate was selected.
-            return 0;
+            return B0;
     }
 }
 
@@ -412,7 +412,7 @@ mraa_uart_set_baudrate(mraa_uart_context dev, unsigned int baud)
 
     // set our baud rates
     speed_t speed = uint2speed(baud);
-    if (speed == 0)
+    if (speed == B0)
     {
         syslog(LOG_ERR, "uart%i: set_baudrate: invalid baudrate: %i", dev->index, baud);
         return MRAA_ERROR_INVALID_PARAMETER;
--- a/src/uart/uart.c
+++ b/src/uart/uart.c
@@ -114,6 +114,54 @@ uint2speed(unsigned int speed)
     }
 }
 
+static unsigned int speed_to_uint(speed_t speedt) {
+    struct baud_table {
+        speed_t speedt;
+        unsigned int baudrate;
+    };
+    static const struct baud_table bauds[] = {
+        { B50, 50 },
+        { B75, 75 },
+        { B110, 110 },
+        { B150, 150 },
+        { B200, 200 },
+        { B300, 300 },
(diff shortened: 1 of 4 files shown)


src/uart/uart.c
```diff
@@ -114,6 +114,54 @@ uint2speed(unsigned int speed)
...
+#if !defined(MSYS)
+        { B3500000, 3500000 },
+        { B4000000, 4000000 },
+#endif
+        { B0, 0} /* Must be last in this table */
+    };
+    int i = 0;
+
+    while (bauds[i].baudrate > 0) {
+        if (speedt == bauds[i].speedt)
```
alext-mkrs: Please use braces for one-line `if` statements as well, to follow the repo style. You can also use `clang-format` and config file from the root of the repo to check/correct style before submission.

arfoll:
1. A little bit of both, I didn't want to clutter the API with things I was unclear if people really needed. Originally we didn't have such flexible/crazy platform support/flexibility so it wasn't so useful. I'd happily merge things for all other buses/io.
2. So you mean nullary functions in C not being declared with 'void' specifying the arity of 0 explicitly? After educating myself a little on the topic it definitely seems that for C it's the better choice (makes no different in C++ apparently). But I seem to remember we had some weird compilation issue because of void being translated weird in SWIG. But it's passed all my tests now so this was probably back when we have SWIG translating the C code rather than the C++ code. So go ahead and use it, we can change all nullary functions to this style.

TechNexion:
Thanks for the clarifications and noticing the style flaws, will adapt my PR.

@arfoll At the moment I don't need the extra methods for other buses/io. It is just a habit that when I write accessor functions, I try to make them "all" in one go. Usually easier that way than going back and add them one by one. And it is always a bit of a "huh?" to find an obvious function missing. 
Whichever is fine with me (push the accessor patches or not).
And no need to go fixing all the incomplete declarations, just accept both ways for now.

arfoll:
Yup agreed, please push the accessor functions if you have them all ready.

PR #740, opened 2017-05-19
Mraa uart

TechNexion:
This PR introduces the mraa-uart tool. It is a tool similar to mraa-gpio, but intended for querying, testing, debugging and using UARTs with the MRAA framework.

Inside are six patches:
1. Fixes a return for uint to speed_t conversion function. Probably changes no functionality.
2. Introduces the inverse function for the above function, converting speed_t to baudrates.
3.+4. Introduces accessor functions for querying the number of uart, spi, pwm, gpio and analogs pins on the platform.
5. Adds a function to query the current settings of a UART. It has two use modes: one by uart index, and one using a devpath. 
   RFC: Is this better this way than having one query function for each parameter?
6. This is the purpose of some of the above patches: a tool to use UARTs with MRAA. The tool can send and receive data, by using MRAA names, devpaths or MRAA indices. It can also query, show and set UART settings. The syntax is intended to be similar to mraa-gpio - run the tool for a brief usage. This tool is hopefully useful for porting new boards to MRAA, since it lessens the burdain to validate UART functionality.

Comments very welcome. Normally this would be an RFC instead of a PR since the tool in patch 6 is more than null-pointer check. I probably messed up on the style standard somewhere too..


arfoll:
This looks good to me, not sure I like the uart settings function but can't think of another way of doing it easily without thousands of small getters so will have to do :)

PR #741, opened 2017-05-20
Static & manual code analysis fixes - the next part

alext-mkrs:
This one marks finish for fixes of things marked as "bugs" in non-example code (there are also "smells", but we'll get to that later).

### DIFF
--- a/src/doxy2swig.py
+++ b/src/doxy2swig.py
@@ -121,6 +121,7 @@ class Doxy2SWIG:
         # ignore pure whitespace
         m = self.space_re.match(txt)
         if m and len(m.group()) == len(txt):
+            # do nothing
             pass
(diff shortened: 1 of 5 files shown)


alext-mkrs:
@arfoll, ping? These are basically trivial and I'd let them in right away by myself, but I just believe that additional review never hurts even if that's just a glance. :)

alext-mkrs:
That's because it's actually not used in that "if" branch or after it, so it was flagged as meaningless. I've left the syslog message as the code actually does set the bus 6 up (we are on an Arduino board in that clause).

arfoll:
Makes sense to me, all merged! Thanks!

alex-ter:
Thanks!

PR #744, opened 2017-05-26
Modify Travis Jobs

dnoliver:
Use Docker and docker-compose to build and run test

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,100 @@
+FROM ubuntu:16.04
+
+ENV DEBIAN_FRONTEND noninteractive
+
+# Update apt-get
+RUN apt-get -y update && \
+  # Install apt-utils
+  apt-get -y --no-install-recommends install apt-utils && \
+  # Main Build Dependencies
+  apt-get -y --no-install-recommends install git build-essential cmake clang-3.8 g++-4.8 wget libpcre3 libpcre3-dev \
+  # Docs Build Dependencies
+  python-sphinx doxygen graphviz \
+  # Python Build Dependencies
+  python python-dev python3 python3-dev \
+  # Java Build Dependencies
+  default-jre default-jdk \
+  # Json Platform Build Dependencies
+  pkg-config libjson0 libjson0-dev
+
+# Swig Build Dependencies
+RUN wget http://iotdk.intel.com/misc/tr/swig-3.0.10.tar.gz && \
+    tar xf swig-3.0.10.tar.gz && cd swig-3.0.10 && \
+    ./configure --prefix=/usr/ && make && make install && cd ..
+
+# Node.js Build Dependencies
+RUN wget -q -O - https://raw.githubusercontent.com/creationix/nvm/v0.33.2/install.sh | bash
+
+# Set Workdir
+WORKDIR /usr/src/app
+
+# Configure Build Arguments
+ARG BUILDARCH
+ARG BUILDDOC
+ARG BUILDSWIG
+ARG BUILDSWIGPYTHON
+ARG BUILDSWIGNODE
+ARG BUILDSWIGJAVA
+ARG USBPLAT=OFF
+ARG FIRMATA=OFF
+ARG ONEWIRE=OFF
+ARG JSONPLAT
+ARG IMRAA=OFF
+ARG FTDI4222=OFF
+ARG IPK=OFF
+ARG RPM=OFF
+ARG ENABLEEXAMPLES
+ARG INSTALLGPIOTOOL=OFF
+ARG INSTALLTOOLS=OFF
+ARG BUILDTESTS
+ARG CC
+ARG CXX
+ARG NODE_VERSION
+
+# Configure Build Environment
+ENV NVM_DIR /root/.nvm
+ENV JAVA_HOME /usr/lib/jvm/java-8-openjdk-amd64/
+ENV CC $CC
+ENV CXX $CXX
+RUN . $NVM_DIR/nvm.sh && nvm install $NODE_VERSION && nvm use $NODE_VERSION && \
+    npm install -g node-gyp && node-gyp install
+
+# Copy sources
+COPY . .
+
+# Update Submodule
+RUN git submodule update --init --recursive
+
+# Fix line ending issue in src/doxy2swig.py and do it executable
+RUN tr -d "\r" < src/doxy2swig.py > src/_doxy2swig.py && \
+    mv src/_doxy2swig.py src/doxy2swig.py && \
+    chmod u+x src/doxy2swig.py
+
+# Change Workdir to build directory
+WORKDIR /usr/src/app/build
+
+# Run cmake
+RUN . $NVM_DIR/nvm.sh && cmake \
+    -DSWIG_EXECUTABLE=/usr/bin/swig \
+    -DSWIG_DIR:PATH=/usr/share/swig/3.0.10/ \
+    -DBUILDARCH=$BUILDARCH \
+    -DBUILDDOC=$BUILDDOC \
+    -DBUILDSWIG=$BUILDSWIG \
+    -DBUILDSWIGPYTHON=$BUILDSWIGPYTHON \
+    -DBUILDSWIGNODE=$BUILDSWIGNODE \
+    -DBUILDSWIGJAVA=$BUILDSWIGJAVA \
+    -DUSBPLAT=$USBPLAT \
+    -DFIRMATA=$FIRMATA \
+    -DONEWIRE=$ONEWIRE \
+    -DJSONPLAT=$JSONPLAT \
+    -DIMRAA=$IMRAA \
+    -DFTDI4222=$FTDI4222 \
+    -DIPK=$IPK \
+    -DRPM=$RPM \
+    -DENABLEEXAMPLES=$ENABLEEXAMPLES \
+    -DINSTALLGPIOTOOL=$INSTALLGPIOTOOL \
+    -DINSTALLTOOLS=$INSTALLTOOLS \
+    -DBUILDTESTS=$BUILDTESTS \
+    ..
+
+CMD make
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -1,0 +1,83 @@
+version: '2'
+
+services:
+
+  main:
+    build:
+      context: .
+      dockerfile: Dockerfile
+      args:
+        - http_proxy
+        - https_proxy
+        - no_proxy
+        - BUILDDOC=OFF
+        - BUILDSWIG=OFF
+        - BUILDSWIGPYTHON=OFF
+        - BUILDSWIGJAVA=OFF
+        - BUILDSWIGNODE=OFF
+        - USBPLAT=OFF
+        - FIRMATA=OFF
+        - ONEWIRE=OFF
+        - JSONPLAT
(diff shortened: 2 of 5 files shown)


alext-mkrs:
Any specific reasoning behind this proposed change?

arfoll:
Generally this looks pretty good, although like @alext-mkrs I'm curious as to why this was done? Also it seems builds with python3 fail which I guess is just because they expect py2?

dnoliver:
@arfoll @alext-mkrs, this PR basically tries to solve the issue where a developer say "it works in my machine, but do not work on CI server". By using the Docker image, the CI environment and the develop/test environment are the same, they use the same toolchain and the same commands to perform a build. So this is trying to normalize the development environment we could all use, and will be used in CI too. 

Before this, the build procedure was coupled to what Travis provide, and replicating the build environment in a development machine mean install all the tools and try to figure out how to execute the build commands. Now Travis just do `docker-compose run ${TARGET}`, and you can do the same in your machine as well :)

About python3, @pylbert told me that he has a solution to the failing test, I am looking forward to hear from him.

Dockerfile
```diff
@@ -71,13 +59,25 @@ ENV CXX $CXX
 RUN . $NVM_DIR/nvm.sh && nvm install $NODE_VERSION && nvm use $NODE_VERSION && \
     npm install -g node-gyp && node-gyp install
 
+# Copy sources
+COPY . .
```
alext-mkrs: I'm probably missing something obvious but why are we copying current dir to current dir?

docker-compose.yaml
```diff
@@ -0,0 +1,83 @@
...
+        - CXX=g++-4.8
+    command: bash -c "make mraajava && make test"
+
+  node4:
+    extends: main
+    build:
+      args:
+        - BUILDSWIG=ON
+        - BUILDSWIGNODE=ON
+        - NODE_VERSION=v4.1.2
```
alext-mkrs: Why this version? I would rather suggest 4.4.x as the LTS one

Dockerfile
dnoliver: This Docker instruction copy from the local file system to the container file system. (https://docs.docker.com/engine/reference/builder/#copy)

in `+WORKDIR /usr/src/app` we tell docker that every command executed from now on will be relative to that folder, so `COPY . .` is equivalent to `COPY . /usr/src/app`

docker-compose.yaml
```diff
@@ -0,0 +1,83 @@
...
+      - http_proxy
+      - https_proxy
+      - no_proxy
+
+  doc:
+    extends: main
+    build:
+      args:
+        - BUILDDOC=ON
+    command: make doc
```
alext-mkrs: This is the only one not wrapped in `bash -c` - is this intentional and why if yes?

docker-compose.yaml
dnoliver: Sounds good to me! I actually choose this version because it was being used in the "old" travis file. It can be easily changed

alext-mkrs:
Thanks for the explanation - makes sense to me and @arfoll - that's more or less along the same lines I was proposing to give container a try earlier (though admittedly I've used more corner case examples of SWIG download problem solving and mock dev env :).

I've added a couple of comments, but generally it looks ok to me - though we'd apparently want to squash all those commits into one or a couple.

docker-compose.yaml
dnoliver: intentional. Docker us `sh` as the default shell to execute commands. `make doc` works fine in `sh`, but i.e. `make mraajava && make test` do not work in `sh` because of the `&&` character. That is why I use `bash -c` for it. It could look better if every command just use `bash -c``everywhere, thought

dnoliver: I think what I said before is not accurate! I could also use `sh -c` to run it! But I have chosen bash just because I like bash more than sh

dnoliver:
@pylbert I am not an expert on CMake, but I think the problem with python3 tests is that is trying to execute those test with python2. If you see the CmakeLists.txt in test/CmakeLists.txt:

```
if (BUILDSWIGPYTHON)
    if (PYTHON2INTERP_FOUND)
        set (PYTHON_DEFAULT_PYTHONPATH "${CMAKE_BINARY_DIR}/src/python/python2")
    elseif (PYTHON3INTERP_FOUND)
        set (PYTHON_DEFAULT_PYTHONPATH "${CMAKE_BINARY_DIR}/src/python/python3")
    endif ()
    if (MOCKPLAT)
        add_subdirectory (mock)
    else()
```

This will always select Python2! because both python2 and python3 is installed in the Docker image.

arfoll:
Looking good!

@dnoliver I fixed it on master 1d37d3b1624ff953beb1d456e2d1a4a3ea1549ec the python testing issue. Btw avoid the merges on your branch by using rebase, it will make it easier to rebase this into one and commit it. I agree with @alext-mkrs please squash all this to one commit when you are ready for me to merge it.

dnoliver:
Just to clarify, are you asking me to create a new PR with all this commits squashed into a single one? If that is the case, I will do:

1. in master branch, `git pull && git checkout -b travis2`
2. in travis2 branch, `git merge --squash travis` to get all this commits squashed into a single one
3. push travis2 branch, Travis will run over dnoliver/travis2 branch, so I can test if python3 tests now pass
4. if CI pass, create a new PR from travis2 branch

is that correct?

arfoll:
no need, just do an interactive git rebase `git rebase -i HEAD~20` (replace 20 with whatever number you have but I like seeing a few...), then squash the relevant commits and force push to this branch. That'll update the PR.

Alternatively create a new branch (avoids having to use reflog/reset to bring back history if you mess up), go back to this branch and do the same thing.

Dockerfile
alex-ter: Ok, makes sense

docker-compose.yaml
```diff
@@ -0,0 +1,83 @@
...
+        - FTDI4222=OFF
+        - IPK=OFF
+        - RPM=OFF
+        - ENABLEEXAMPLES=ON
+        - INSTALLGPIOTOOL=OFF
+        - INSTALLTOOLS=OFF
+        - BUILDTESTS=ON
+        - CC=clang-3.8
+        - CXX=clang++-3.8
+        - NODE_VERSION=v4.1.2
```
alex-ter: This line still has the old Node.js - let's switch to 4.4.x here as well.

Dockerfile
alext-mkrs: Ok, makes sense, thanks

docker-compose.yaml
alext-mkrs: This one still has the old Node version - let's switch to 4.4.x here as well.

docker-compose.yaml
alext-mkrs: > It could look better if every command just use bash -c everywhere, thought

+1 for that, would avoid unnecessary confusion several months from now when no one remembers why everything has `bash -c` and this one doesn't :)

docker-compose.yaml
dnoliver: Fixed in squashed commit

dnoliver:
@arfoll I completed the squash. I never did that with git this was my first time, we learn something new every day :)

dnoliver:
I tested you fix @arfoll. I merged your commit into this branch and here is the result:

```
dnoliver@dnoliver-vm:~/Git/mraa$ docker-compose run python3
Scanning dependencies of target mraa
[  4%] Building C object src/CMakeFiles/mraa.dir/mraa.c.o
/usr/src/app/src/mraa.c:880:28: warning: comparison of array
      'plat->pins[i].name' not equal to a null pointer is always true
      [-Wtautological-pointer-compare]
         if (plat->pins[i].name != NULL &&
             ~~~~~~~~~~~~~~^~~~    ~~~~
1 warning generated.
[  8%] Building C object src/CMakeFiles/mraa.dir/gpio/gpio.c.o
[ 12%] Building C object src/CMakeFiles/mraa.dir/i2c/i2c.c.o
[ 16%] Building C object src/CMakeFiles/mraa.dir/pwm/pwm.c.o
[ 20%] Building C object src/CMakeFiles/mraa.dir/spi/spi.c.o
[ 24%] Building C object src/CMakeFiles/mraa.dir/aio/aio.c.o
[ 28%] Building C object src/CMakeFiles/mraa.dir/uart/uart.c.o
[ 32%] Building C object src/CMakeFiles/mraa.dir/iio/iio.c.o
[ 36%] Building C object src/CMakeFiles/mraa.dir/x86/x86.c.o
[ 40%] Building C object src/CMakeFiles/mraa.dir/x86/intel_galileo_rev_d.c.o
[ 44%] Building C object src/CMakeFiles/mraa.dir/x86/intel_galileo_rev_g.c.o
[ 48%] Building C object src/CMakeFiles/mraa.dir/x86/intel_edison_fab_c.c.o
[ 52%] Building C object src/CMakeFiles/mraa.dir/x86/intel_de3815.c.o
[ 56%] Building C object src/CMakeFiles/mraa.dir/x86/intel_nuc5.c.o
[ 60%] Building C object src/CMakeFiles/mraa.dir/x86/intel_minnow_byt_compatible.c.o
[ 64%] Building C object src/CMakeFiles/mraa.dir/x86/intel_sofia_3gr.c.o
[ 68%] Building C object src/CMakeFiles/mraa.dir/x86/intel_cherryhills.c.o
[ 72%] Building C object src/CMakeFiles/mraa.dir/x86/up.c.o
[ 76%] Building C object src/CMakeFiles/mraa.dir/x86/intel_joule_expansion.c.o
[ 80%] Building C object src/CMakeFiles/mraa.dir/version.c.o
[ 84%] Linking C shared library libmraa.so
[ 84%] Built target mraa
[ 88%] Swig source
Scanning dependencies of target _python3-mraa
[ 92%] Building CXX object src/python/python3/CMakeFiles/_python3-mraa.dir/mraa3PYTHON_wrap.cxx.o
[ 96%] Building C object src/python/python3/CMakeFiles/_python3-mraa.dir/__/mraapy.c.o
[100%] Linking CXX shared module _mraa.so
[100%] Built target _python3-mraa
Test project /usr/src/app/build
    Start 1: py_general
libmraa[160]: libmraa version v1.7.0-27-g2b79330 initialised by user 'root' with EUID 0
libmraa[160]: Platform not supported, not initialising
libmraa[160]: libmraa initialised for platform 'Unknown platform' of type 98
1/3 Test #1: py_general .......................   Passed    0.23 sec
    Start 2: py_platform
libmraa[161]: libmraa version v1.7.0-27-g2b79330 initialised by user 'root' with EUID 0
libmraa[161]: Platform not supported, not initialising
libmraa[161]: libmraa initialised for platform 'Unknown platform' of type 98
2/3 Test #2: py_platform ......................   Passed    0.02 sec
    Start 3: py_gpio
libmraa[162]: libmraa version v1.7.0-27-g2b79330 initialised by user 'root' with EUID 0
libmraa[162]: Platform not supported, not initialising
libmraa[162]: libmraa initialised for platform 'Unknown platform' of type 98
3/3 Test #3: py_gpio ..........................   Passed    0.02 sec

100% tests passed, 0 tests failed out of 3

Total Test time (real) =   0.28 sec
```

With Mock platform also works!

arfoll:
Merged. Thanks alot!

PR #747, opened 2017-05-30
Fixing Beaglebone module to work with kernel 4.x.

nrcrast:
Made a few issues under my work account, @ncrastanaren.

* No more loading of individual peripheral capes. In 4.x, it's preferred to use a universal cape and configure the pins during initialization.
* PWM uses pwmchipX interface
* General cleanup of pin definitions. Created macro BUILD_PIN that's a little bit easier to look at, so you don't have to do the mental math of converting, for example, pin 76 -> P9_30

Signed-off-by: Nick Crast <nrcrast@gmail.com>

### DIFF
--- a/src/arm/beaglebone.c
+++ b/src/arm/beaglebone.c
@@ -23,54 +23,196 @@
  * WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
  */
 
+#include <glob.h>
+#include <mraa/types.h>
 #include <stdlib.h>
 #include <string.h>
 #include <sys/mman.h>
-#include <mraa/types.h>
 
 #include "common.h"
 #include "arm/beaglebone.h"
 #include "arm/am335x.h"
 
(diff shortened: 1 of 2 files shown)


nrcrast:
Tested: I2C, SPI, PWM

arfoll:
@svlstriker can you take a look at this? I'd like to merge this but don't want to break the phyboard stuff either.

nrcrast:
UART was also broken. At some point the higher level uart interface started using uart_dev->device_path, which isn't set in the beaglebone module. Fixed in latest pair of commits.

nrcrast:
There also seems to be a problem with the ADC that I am chasing down. You have some math in aio.c:
`pin = aio + board->gpio_count;`

This doesn't really work with the BBB, the gpio_count is currently 45, but looking at the pinout the AIN pins start at 81. 

Why doesn't this function just accept the raw pin like others do?

arfoll:
Closed as #750 contains this fix.

PR #750, opened 2017-05-31
Beaglebone Black AIO Fix

nrcrast:
@arfoll Here is my attempt at fixing the AIO logic for the BBB. Sorry, my fixes are lumped in with the other BBB stuff, but check out the latest commit. 

**Commit message**

Beaglebone AIO pins seem to be a little different than most boards, so this is my attempt to work with that without impacting other boards. I added a new flag in mraa_board_t to indicate whether or not the aio pins are sequential. On the beaglebone, they are not. To go along with this, I added a new device mraa_aio_dev_t, that will map each aio to a physical pin.

In the main aio logic, if aio_non_seq is true for the board, the manual mapping is used, otherwise the old mathematical mapping is used.

### DIFF
--- a/include/mraa_internal_types.h
+++ b/include/mraa_internal_types.h
@@ -43,6 +43,7 @@
 // Max count for various busses
 #define MAX_I2C_BUS_COUNT 12
 #define MAX_SPI_BUS_COUNT 12
+#define MAX_AIO_COUNT 7
 #define MAX_UART_COUNT 6
 #define MAX_PWM_COUNT 6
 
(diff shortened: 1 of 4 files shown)


nrcrast:
This fix does assume that all of the board initialize the mraa_board_t structure correctly. Might be worth having a function that allocates an mraa_board_t, initializes it properly to 0's/NULL as appropriate, and returns a pointer. I didn't check the other boards to see if they did `calloc` to allocate the struct like the beaglebone module does.

Propanu:
Looks good to me. Although I think we should update the documentation for the AIO functions/class and mention analog pins are always 0 indexed regardless of their position on the I/O header.

nrcrast:
@Propanu I agree with that, for sure. Before looking at the source, I thought that the mraa_aio_init function took the raw pin number as an argument. Looking at the docs again, they do say channel number, but the variable is called `pin`, so it's a little confusing.

arfoll:
Merged, this is awesome. Thanks alot

arfoll:
Updated the docs as @Propanu suggested

PR #755, opened 2017-06-05
The last portion of static code analysis fixes - this time in examples

alext-mkrs:
After getting this one in we should be down to zero (counting those meaningful ones) and next thing I'll look at would be implementing automated scans within Travis.


arfoll:
Merged! Big thanks

PR #756, opened 2017-06-06
Fix Travis execution for gcc

dnoliver:
Travis was only executing jobs with clang compiler, because the CC and CXX value was hardcoded in the docker-compose.yaml file. This new syntax allow to load the value from the execution environment, or set a default if no value is found

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -45,6 +45,12 @@ matrix:
 services:
   - docker
 
+before_install:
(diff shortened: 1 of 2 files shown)


alext-mkrs:
+1, looks good to me, we indeed were running only clang-based builds for some time.

alext-mkrs:
Squashed and committed with 000746c20d15da51e4c32b618888a3ab79be646c

Issue #761, opened 2017-06-15
Platform loading Segmentation Fault


nigeltiany:
I am trying to platform load the Odroid XU4. 
Here is my odroid.json
```json
{
    "version": "0.0.1",
    "platform": [
        {
            "id": 1010,
            "name": "Odroid",
            "pin_count": 32, // 30 GPIO Plus 2 Analog inputs
            "gpio_count": 30,
            "aio_count": 2,
            "uart_count":1,
            "i2c_count": 4,
            "spi_count": 1,
            "pwmDefPeriod": 0,
            "pwmMaxPeriod": 0,
            "pwmMinPeriod": 0
        }
    ],
    "layout": [
        {"label": "AIN-0"}, // AIO index 0 lable
        {"label": "GPIO-4"},
        {"label": "GPIO-5"},
        {"label": "GPIO-6"},
        {"label": "GPIO-7"},
        {"label": "GPIO-8"},
        {"label": "GPIO-9"},
        {"label": "GPIO-10"},
        {"label": "GPIO-11"},
        {"label": "GPIO-13"},
        {"label": "GPIO-14"},
        {"label": "GPIO-15"},
        {"label": "GPIO-16"},
        {"label": "GPIO-17"},
        {"label": "GPIO-18"},
        {"label": "GPIO-19"},
        {"label": "GPIO-20"},
        {"label": "GPIO-21"},
        {"label": "GPIO-22"},
        {"label": "AIN-3"}, // AIO index 19 label
        {"label": "GPIO-24"},
        {"label": "GPIO-25"},
        {"label": "GPIO-26"},
        {"label": "GPIO-27"},
        {"label": "X-GPIO-4"},
        {"label": "X-GPIO-5"},
        {"label": "X-GPIO-6"},
        {"label": "X-GPIO-7"},
        {"label": "X-GPIO-9"},
        {"label": "X-GPIO-10"},
        {"label": "X-GPIO-11"},
        {"label": "X-GPIO-12"}
    ],
    "AIO": [
        { "index": 0, "rawpin": "in_voltage0_raw"},
        { "index": 19, "rawpin": "in_voltage3_raw"}
    ],
    "GPIO": [
        { "index": 1, "rawpin": 173 },
        { "index": 2, "rawpin": 174 },
        { "index": 3, "rawpin": 171 },
        { "index": 4, "rawpin": 192 },
        { "index": 5, "rawpin": 172 },
        { "index": 6, "rawpin": 191 },
[...]

PR #763, opened 2017-06-15
jsonplatform.c: fix potential segfault at pin label processing

### LINKED ISSUE #738: jsonplatform.c: potential pin label truncation
While analyzing SonarQube findings for mraa I've come across an interesting place (not flagged by Sonar, BTW), namely [line 217 of `jsonplatform.c`](https://github.com/intel-iot-devkit/mraa/blob/master/src/json/jsonplatform.c#L217),

```c
temp_string = json_object_get_string(jobj_temp);
// set the gpio label
strncpy(board->pins[pos].name, temp_string, 8);
```

where we seem to do several suspicious things at once:

1. Copy exactly 8 symbols, while `name` in `mraa_pininfo_t` is declared as `name[MRAA_PIN_NAME_SIZE]` (for the non-PeripheralManager case) where `MRAA_PIN_NAME_SIZE` equals 12. There is no mention in the docs of what the max length of the label is, so we potentially truncate otherwise valid labels (9-12 symbols) and don't null-terminate all ofthem properly;
1. For PeripheralManager `.name` is declared as `char *name` and we don't seem to allocate the memory for it anywhere before copying. I'm not sure if this codepath is at all relevant for AT?

@Hbrinj could you please help me understand if there's any specific idea behind those or we need to fix that?

alext-mkrs:
Also update docs to reflect the max label length.

Fixes #738.

### DIFF
--- a/docs/jsonplatform.md
+++ b/docs/jsonplatform.md
@@ -41,7 +41,7 @@ The number of json objects in layout should be equal to the pin_count.
 
 |Key        |Type   |Required   |Description                                    |
 |-----------|-------|-----------|-----------------------------------------------|
-|label      |string |yes        | The label used to describe the pin            |
+|label      |string |yes        | The label used to describe the pin (11 characters max) |
 |invalid    |boolean|no         | Sets the labeled pin as an invalid pin        |
 
 ### GPIO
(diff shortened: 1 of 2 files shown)


alext-mkrs:
May also potentially fix #761

arfoll:
Merged!

PR #765, opened 2017-06-17
mraa.c/x86.c: use exact match in strncmp() to avoid surprises

### LINKED ISSUE #736: Refactor strncmp()-using statements to do exact matches instead of substring ones
In our string comparisons we universally use `strncmp()`, but frequently with a length parameter not accounting for a null-terminator, causing a partial match to be used, so this may bite us at some point (e.g. "gpio1" would match successfully against "gpio12").

Lets sweep those and make sure we do exact matches.

alext-mkrs:
Closes #736.


arfoll:
Merged!

PR #768, opened 2017-06-21
gpio : Add support for input pull up/down modes

Mani-Sadhasivam:
Make use of 'active_low' interface in sysfs for configuring input pin
in pull up / pull down mode. C++ binding also has been added.


arfoll:
Merged! Thanks alot

PR #772, opened 2017-06-24
x86.c: sanitize data read from sysfs during board detection

### LINKED ISSUE #771: Latests commits break openaps on edison with ValueError: Invalid GPIO pin specified error
Multiple users of openaps reported that the latest master checkout causes
`ValueError: Invalid GPIO pin specified` with openaps. Previous master checkouts didn't have that error.

Latest comit that was ok:
https://github.com/intel-iot-devkit/mraa/commit/e0052acfa7ab441b5417a0fbd010ce77b285a4d8

So the problem was caused by one of the following commits (or the combination):
- https://github.com/intel-iot-devkit/mraa/commit/30bbb886850d2a7c55c5803e5c830e5c6cfd2e1a
- https://github.com/intel-iot-devkit/mraa/commit/d544e3c2f60ff3dd1b2501fb6e8b4136347675f3

alext-mkrs:
Fixes #771. Tested on Edison and mock (though the latter isn't affected).

### DIFF
--- a/src/x86/x86.c
+++ b/src/x86/x86.c
@@ -51,6 +51,9 @@ mraa_x86_platform()
     FILE* fh = fopen("/sys/devices/virtual/dmi/id/board_name", "r");
     if (fh != NULL) {
         if (getline(&line, &len, fh) != -1) {
+            // Sanitize input by terminating at any of possible end of line chars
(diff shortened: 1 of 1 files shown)


arfoll:
Merged. Thanks for the quick fix!

PR #779, opened 2017-07-01
jsonplatform.c: properly reallocate memory for platform name

alext-mkrs:
Instead of copying the string value we assigned pointers directly, causing double free in `mraa_deinit()` when JSON platform is used. One is done on `plat->platform_name`, another one - on `platform_name`, not sure why we need both BTW. Normally they are pointing to different buffers, but in this case they didn't.

I could use `realloc()`, but I like the zeroing `calloc()` does automatically, thus a `free()` + `calloc()` pair.


arfoll:
Makes good sense. Merged.

PR #780, opened 2017-07-02
intel_edison_fab: enable support for vanilla kernels >= 4 for serial

htot:
mraa: patch for vanilla kernel >= 4.x.x

Vanilla kernel have the serial port on ttyS1 instead of ttyMFD1 and don't support changing pinmux at run time. It is envisioned that pinmux should be set by acpi tables that will be passed from u-boot to the kernel. 
For now serial port pinmux is pre-configured to allow ttyS1 with flow control. I.e. J17-14, J18-12, J18-13 and J19-8 on mini-breakout board and IO0, IO1, IO2 and IO4 on the arduino board are reserved for serial communication.

This patch detects the kernel version and sets the default serial port to ttyS1 for kernels >= 4.x.x and disables pinmux changes. A warning is set to syslog when this attempted.

Signed-off-by: Ferry Toth <ftoth@exalondelft.nl>

My branch eds is from master tag v1.7.0

### DIFF
--- a/src/x86/intel_edison_fab_c.c
+++ b/src/x86/intel_edison_fab_c.c
@@ -31,6 +31,7 @@
 #include <linux/spi/spidev.h>
 #include <errno.h>
 #include <time.h>
+#include <sys/utsname.h>
 
 #include "common.h"
 #include "x86/intel_edison_fab_c.h"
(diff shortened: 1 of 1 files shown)


arfoll:
Sorry about the slow reply - this looks good, I've rebased it and put it on the edison branch for now, I want to run a few more tests on it and then I'll merge.

arfoll:
Merged, thanks!

Issue #782, opened 2017-07-03
Prepare docs on static code analysis


alext-mkrs:
Just to have it gathered in one place and explicit - let's create some docs. This is a note to myself to do that.

We want
1. instructions on configuring automated scans in Travis+GitHub+SonarCloud set
1. instructions on how to run a scan manually outside of CI system

PR #785, opened 2017-07-04
Travis + Docker improvements

dnoliver:
This changes are intended to improve the current build infrastructure with Docker in Travis. Previously, there was a single Docker image that contains every build tool for C/C++, Node, Java, and Python. this make each Travis builder to install unnecessary packages to build a single target. Adding Android dependencies, made that image a ~4.5Gb image, so I tried to change that.

Now there is a Dockerfile hierarchy. There is a "base" image that provides C/C++ build dependencies, and other common dependencies. The "python" image install python2 and python3. The "node" image install an specific Node.js version. And the "java" image install jdk and jre. The build process is now a little bit more complex (you need to build the "base" image first, and then the other ones), but the final image size and build times are improved, which is nice :). All the Dockerfiles are now located in the "docker" folder.

There is an important change over there: using volumes. Previously, docker build copied all the sources from the host to the image, and then build the targets using the sources in the image. Now it is using a volume to mount all the sources into the running container. This has the following results:
1. The final image size is reduced, the build is faster
2. there is no image rebuild when the sources changes! it just mount the sources from the host to the container and use them.
3. the "build" folder is also available in the host file system, so you don't need to re-compile everything from scratch. Make decide which file is changed and only recompile what its needs
4. We can publish those images, and use them as ready-to-use build environments. 

On top of this changes, I implemented the Android Thing build. The "android" image install the NDK and Android Things Native lib (which adds ~3Gb to the final image) and use the "scripts/build-android.sh" command to generate aar packages. 

Travis is already building Android Thing packages, but I am not sure if those packages work or not. I don't have the infrastructure to test aar packages.

Additionally, I added a missing FTDI4222 build dependencies, so now FTDI4222 works. 

@bjbeare @pylbert  FYI this add Android Thing Build for mraa
@Propanu If you think this Docker hierarchy is a good idea, we should do the same for upm


bjbeare:
Thanks. I'll give this a go and we'll test the resulting AAR files.
It may have to wait till  tonight.. I've been unsuccessful (so far) getting docker-compose to build the image behind the firewall.

dnoliver:
@alext-mkrs I had some conflicts with your changes about Sonar. I made some changes to resolve them:

1. I moved the sonar-scan.sh script to the script folder, 
2. sonar-scan target now depends from the "base" image. This mean that it does not have node.js, java, and python installed. So I turned off the Swig options to build those targets. I would like to take a look at what Sonar do in the future. Do you have any example report to share?

alext-mkrs:
@dnoliver, moving to a separate dir is not a problem, but disabling swig and the rest is a bad idea, because that decreases scan coverage (it scans what's being compiled and there are some ifdef'd pieces). We want to have all options maxed out for a scan.

You can see reports in the SonarCloud org, see links in the PR introducing this.

dnoliver:
@alext-mkrs you are right, it is a bad idea. a possible solution is this: sonar is now installed on the base image, so every other image has it available. i.e. I can run sonar-scan with "mraa-python" to scan the python sources, and the same for java and node.js

I can run sonar 3 times, or we can build all the targets first in the build directory and run sonar then. what do you think?

bjbeare:
Please add the following to the documentation before you merge.

1. To run docker behind a proxy, you must first configure docker.io to use the proxy. The instructions for that are found at: https://stackoverflow.com/questions/23111631/cannot-download-docker-images-behind-a-proxy

2. To have your docker client image use the proxy, you must then set some environment variables before running docker-compose. You *must* use the IP address (number) not the DNS name for these. For example, for my company proxy, I use:
export http_proxy=http://10.22.230.62:911
export https_proxy=http://10.22.230.62:912
export no_proxy=127.0.0.1,localhost

bjbeare:
Please all the following to the documentation before you merge:

The image is built in stages and there are not (currently) dependencies between the stages. To build the Android *.aar files, you should:

arfoll:
I'm not sure teaching users how to set proxies in Linux is *our* job ;).
Theres nothing there that is out of the ordinary is there?

Btw you should look at redsocks it'll simplify your life when dealing with
proxies that shall remain nameless!

bjbeare:
I hear you... but I wasted a great deal of time figuring this out and it isn't unreasonable to have a .md file to get people pointed in  the right direction.

arfoll:
I can definately empathise with time wasted on proxies :/

Sorry about the bikeshed :) obv nothing wrong with a few more instruction.

dnoliver:
According to @bjbeare test, the aar package generated with Docker works! :D

docs/building.md
```diff
@@ -199,3 +199,173 @@ added to the CMAKE_MODULE_PATH.
...
+
+```sh
+# Build mraa documentation
+$ docker-compose run doc
+# Build mraa python2 package and run python2 tests
+$ docker-compose run python2
+# Build mraa python3 package and run python3 tests
+$ docker-compose run python3
+# Build mraa java package and run java tests
+$ docker-compose build java
```
alext-mkrs: Shouldn't this be "run java"?

docs/building.md
```diff
@@ -199,3 +199,173 @@ added to the CMAKE_MODULE_PATH.
...
+
+To improve build times and images sizes, mraa use a build hierarchy to incrementally
+create a build environment. Find below a brief description of them:
+
+1. **mraa-base:** Provides the basic infrastructure and tools to compile C/C++ code and documentation.
+2. **mraa-python:** Provides the python2/python3 build tools. Depends on `mraa-base`.
+3. **mraa-java:** Provides the Java build tools. Depends on `mraa-base`.
+4. **mraa-android:** Provides the Android Things build tools.  Depends on `mraa-java`.
+5. **mraa-node4:** Provides the Node.js v4.4.7 build tools. Depends on `mraa-base`.
+6. **mraa-node5:** Same as `mraa-node4`, but using Node.js v5.12.0.
```
alext-mkrs: Let's add the `sonar-scan` target description as well, for completeness (the  inner workings would be decribed in a different document file, I - or someone else - would create within #782)

dnoliver: Done in https://github.com/intel-iot-devkit/mraa/pull/785/commits/6119f4478481d0941ed033e0d89db3f8b71cf06b

docs/building.md
dnoliver: Fixed in https://github.com/intel-iot-devkit/mraa/pull/785/commits/6119f4478481d0941ed033e0d89db3f8b71cf06b

docs/building.md
alext-mkrs: Thanks for adding that, in general the change looks ok to me by inspection, but I also plan to do some more detailed tests later this week and will post final review after that.

dnoliver: @alext-mkrs I did a rebase of this branch with master. hope I did it well

dnoliver:
@alext-mkrs I made a [test pr](https://github.com/dnoliver/mraa/pull/1) to check sonar-scan execution, and it can run properly. But I can see several issues, check here https://github.com/dnoliver/mraa/pull/1/files/0bafa62e1477d5ebd003dcf17126496e17164bde#issuecomment-315118267

@Propanu you disabled sonar scan on every fork in this commit https://github.com/intel-iot-devkit/mraa/commit/4f3eae90e8adb66f4a66aafbd2fe661bfa9e12d2. I think that is not correct! isn't it better to run sonar scan only if the required settings are present? if there is no SONAR_ORG SONAR_PROJ_KEY SONAR_TOKEN GITHUB_TOKEN env var, just do not run sonar.

docker/Dockerfile.base
```diff
@@ -0,0 +1,41 @@
...
+WORKDIR /opt
+
+# Swig Build Dependencies
+RUN wget -q http://iotdk.intel.com/misc/tr/swig-3.0.10.tar.gz && \
+    tar xf swig-3.0.10.tar.gz && cd swig-3.0.10 && \
+    ./configure --prefix=/usr/ && make -j8 && make -j8 install && cd .. && \
+    rm -fr /opt/swig-3.0.10.tar.gz
+
+# FTDI4222 Build Dependencies
+RUN mkdir libft4222-linux-1.3.1.117 && cd libft4222-linux-1.3.1.117 && \
```
alext-mkrs: while we are at it - it would be worth it to split out the version string into a separate var, so upon update one doesn't have to update in zillion places [and surely miss one of them] :)

docker/Dockerfile.base
```diff
@@ -0,0 +1,41 @@
...
+  # Imraa Platform Build Dependencies
+  libudev-dev \
+  # Doc Build Dependencies
+  python-sphinx doxygen graphviz
+
+# Set library download folder to /opt
+WORKDIR /opt
+
+# Swig Build Dependencies
+RUN wget -q http://iotdk.intel.com/misc/tr/swig-3.0.10.tar.gz && \
```
alext-mkrs: while we are at it - it would be worth it to split out the version string into a separate var, so upon update one doesn't have to update in zillion places [and surely miss one of them] :)

scripts/sonar-scan.sh
```diff
@@ -22,6 +22,12 @@
 # docker-compose ones are:
 #  - MRAA_SRC_DIR - path to mraa's git clone in the Docker container.
 
+# Check required environment variables and exit if they are not set
+MRAA_SRC_DIR=${MRAA_SRC_DIR:?value not provided}
+SONAR_PROJ_KEY=${SONAR_PROJ_KEY:?value not provided}
+SONAR_ORG=${SONAR_ORG:?value not provided}
+SONAR_TOKEN=${SONAR_TOKEN:?value not provided}
+
```
dnoliver: @Propanu @alext-mkrs changed the sonar-scan.sh to only run if the environment has the proper variables. It will exit with code 1 and print a message to stderr if those env vars are not provided

For reference: http://www.gnu.org/software/bash/manual/bash.html#Shell-Parameter-Expansion

docker/Dockerfile.base
dnoliver: Done in https://github.com/intel-iot-devkit/mraa/pull/785/commits/bae96d8de4165d8a2e33ba62d4e7ed7d9b4f5c35

docker/Dockerfile.base
dnoliver: Done in https://github.com/intel-iot-devkit/mraa/pull/785/commits/bae96d8de4165d8a2e33ba62d4e7ed7d9b4f5c35

Propanu:
@dnoliver @alext-mkrs thanks for pointing this out and reverting the patch. I agree with allowing the sonar scans to fail instead. I've messed around with it a bit more since, and wish they had better integration with Github/Travis (e.g. read org repos that you have access to and allow those scans).

dnoliver:
@Propanu @alext-mkrs I think this is ready to merge!

alext-mkrs:
Cool, I'll take a look tomorrow, ran out of time today.

scripts/sonar-scan.sh
alext-mkrs: +1 LGTM. One nitpick-level suggestion would be to add a variable name to the message to make it easier to detect what's missing, but that's really optional.

alext-mkrs:
So this looks good to me now. I planned to push this myself, but if you want you can do that, now that you've got permissions. The only thing you'll need to do before that is squash all those 16 commits into 1 before committing as this is really just one logical change. Let me know if you want me to squash and push, that's not a problem and I can do that easily.

scripts/sonar-scan.sh
dnoliver: This is the error message you will have if there is no value on it, the variable name and line number is included:

```
dnoliver@dnoliver-vm:~/Git/mraa$ ./scripts/sonar-scan.sh
./scripts/sonar-scan.sh: line 26: MRAA_SRC_DIR: value not provided
```

dnoliver:
@alext-mkrs according to github documentation here https://help.github.com/articles/configuring-commit-squashing-for-pull-requests/, Github can squash and merge PR into a single commit. But we would need to enable that into this repo (and others). Do you think it will be useful?

alext-mkrs:
I haven't used this feature preferring more control over what's going on, but reading the docs it seems to me that there's either an option of squashing + merging (which might create a merge commit for PRs not rebased to the current master, like this one - and we don't want these, though in a different document they mention the ff merge in this case - so looks like only test will tell 😄) or rebase+merge, which would avoid merge commit, but would not squash PR commits. So it would seem like you can't have both and that's why I'd rather go with a traditional approach.

However if you're up to testing that squashing stuff on a fork, to make sure it doesn't create a merge commit in a situation like this PR, that could be useful.

dnoliver:
ok, will have to test it in a different repo to see how it works. I will do this by hand.

So should I do a `git rebase` to squash all this commits into a single one as @arfoll recommended me in https://github.com/intel-iot-devkit/mraa/pull/744#issuecomment-304939579?

or doing a `git merge --squash` on the master branch will do the job?

Sorry, first time merging something into this repo! :D

arfoll:
You don't merge things on the repo, I do ;-) We don't use merge on this repo to keep a clean history, so you may get asked to rebase your pull request but usually I figure it out. In this case, you're the only one really touching those files so that applies cleanly without any issues, the command I use for reference would be `git am -s 785.patch` (you can do it by checking ou thte fork etc but it's more effort I find) or let gh automagically do it which well... I'm not that trusting!

I split this into 3 commits, not sure it's split all that beautifully but it's fairly intertwined so I'll stop there. Hopefully I didn't break it.

Thanks!

dnoliver:
Travis on master failed because of an sporadic Travis error (it could not fetch docker-compose :( ) I manually re executed the job again

Propanu:
That's perfectly fine, I do that too since it happens all the time, network timeouts and such...

Propanu:
I agree with @arfoll, don't use the Git auto-merge feature, but rather the command line. Rebase your branch on the latest master and squash commits if needed. Then you can switch back to master and merge from there.

PR #789, opened 2017-07-11
FIRMATA: Handling PWM period

malikabhi05:
Signed-off-by: Abhishek Malik <abhishek.malik@intel.com>

### DIFF
--- a/src/pwm/pwm.c
+++ b/src/pwm/pwm.c
@@ -61,10 +61,7 @@ mraa_pwm_write_period(mraa_pwm_context dev, int period)
         mraa_result_t result = dev->advance_func->pwm_period_replace(dev, period);
         if (result == MRAA_SUCCESS) {
             dev->period = period;
-        } else if (result == MRAA_ERROR_FEATURE_NOT_IMPLEMENTED) {
-            return MRAA_SUCCESS;
         }
-
         return result;
     }
     char bu[MAX_SIZE];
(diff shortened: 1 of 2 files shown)


src/pwm/pwm.c
```diff
@@ -61,7 +61,10 @@ mraa_pwm_write_period(mraa_pwm_context dev, int period)
         mraa_result_t result = dev->advance_func->pwm_period_replace(dev, period);
         if (result == MRAA_SUCCESS) {
             dev->period = period;
+        } else if (result == MRAA_ERROR_FEATURE_NOT_IMPLEMENTED) {
+            return MRAA_SUCCESS;
```
malikabhi05: I do not like the idea of doing it this way here, ie returning MRAA_SUCCESS for MRAA_ERROR_FEATURE_NOT_IMPLEMENTED. The other way would be to handle this in the UPM driver, but not sure if that is a good way to go about it either.

arfoll: The return here should be MRAA_ERROR_FEATURE_NOT_IMPLEMENTED passed by pwm.c, I don't see how anything else is appropriate. UPM drivers trying to use this with firmata will have to check/handle it.

arfoll:
I thought we discussed this? MRAA_ERROR_FEATURE_NOT_IMPLEMENTED should never end up being changed to MRAA_SUCCESS. This is a massive hack and shouldn't be in mraa.

malikabhi05:
@arfoll yes we did discuss this and there is a commit which i added "PWM: Removing incorrect MRAA_SUCCESS return" 403281821f22f22fa7fc77f66f7eead88bfb4e1d which actually removes that and its right there. Not sure why you closed the request even though I made the changes.

Propanu:
@arfoll looks like it was changed to match your suggestion, no?

arfoll:
Indeed - I obviously got confused with the adding & removal. Sorry! Next time just rebase and squash the two patches together so it's obvious what you're doing.

Merged.

PR #792, opened 2017-07-19
PWM: Fix crash if pin == phy_pin_count

facchinm:
@arfoll


facchinm:
The problem is not at platform level but at user code level; in fact, if you try calling `mraa_pwm_init(phy_pin_count)` you are writing outside the b->pins array :wink:

arfoll:
Not sure what I was talking about - this makes total sense. Thx & Merged!

PR #794, opened 2017-07-27
Minor fixes for beaglebone I2C initialization

TechNexion:
The i2c_init_pre function for beaglebone seems to be a ... bit messy.
This is an attempt to fix it.
The only functionality change is that on failure to mux bus pins, the function now returns error.
Before these patches the return value is always success, but the way the code was written, that appears to be non-intentional.


arfoll:
Looks all very logical to me! THanks for the patches. merged

PR #795, opened 2017-08-02
Improve Travis build jobs 

dnoliver:
This Pull Request is aimed to lower build times and expand the build matrix in Travis Builds. 

**Docker Images**
* Dockerfiles where moved to https://github.com/dnoliver/iot-devkit-docker
* Docker images are automatically published in https://hub.docker.com/r/dnoliver/  using Travis Deploy
* DockerHub also builds master branch of that repo using [Automated Builds](https://docs.docker.com/docker-hub/builds/)
* In this repo, Travis just pull the already built image, and use it to perform the build.

I hope that we can use an organization instead of my personal dockerhub account to store this images in the future. probably we can use this https://hub.docker.com/u/inteliotdevkit/

**Build Matrix**
* Use build stages to organize jobs in groups (Clang 3.8, Gcc 5, Gcc 6, Additional Jobs)
* Now Travis also build mraa with Gcc 6 and Node.js 6.x
* Add additional jobs: usbplat, firmata, imraa, ftdi4442, ipk, rpm
* ipk is failling because cpack throws a segfault

**Documentation**
* Travis now run a single "doc" job that generates documentation for all target languages
* Several doxygen comments has been modified and added to reduce the big number of warnings when build documentation
* Fixed examples path and page references
* Python Doc now display examples and include doc.i files 
* Added https://github.com/intel-iot-devkit/doxyport as a submodule, and use it to generate java documentation on travis
* Use https://github.com/intel-iot-devkit/doxygen2jsdoc instead of  https://github.com/arfoll/doxygen2jsdoc
* Java, Python, and Node.js documentation still need some improvements, thought 

**Related Pull Requests and changes**
* https://github.com/intel-iot-devkit/doxyport/pull/7
* https://github.com/intel-iot-devkit/doxygen2jsdoc/pull/1
* https://github.com/dnoliver/iot-devkit-docker


dnoliver:
[mraa-doc.tar.gz](https://github.com/intel-iot-devkit/mraa/files/1208325/mraa-doc.tar.gz)

Preview of generated documentation. You can compare it with the latest master one in https://iotdk.intel.com/docs/master/mraa/

dnoliver:
I will merge this into master when Travis  give me green light

.travis.yml
```diff
@@ -60,14 +12,90 @@ before_install:
...
-  - docker-compose build base
-  - if [ "$TARGET" == "android" ]; then docker-compose build java; fi
-  - docker-compose build ${TARGET}
-
-script:
-  - docker-compose run ${TARGET}
+jobs:
+  fast_finish: true
+  allow_failures:
+    - env: TARGET=sonar-scan
```
alext-mkrs: Could you please elaborate why this is set to allow failures? We generally don't want to have an unscanned code in,

dnoliver: Sonar env vars are stored in our Travis CI settings page. Those settings are not shared with external PRs, for security reasons. The sonar-scan.sh script check if those variables exists and fails if they don't. So for a PR, or for a fork without the correct sonar scan configuration, this will fail. 

So, this is an allowed failure to prevent having failed builds for every external PR and fork

Propanu: The other option was what I implemented in the script, to check what fork we're on and kick off the sonar scan based on that. But that breaks scans for those who run them with their own credentials, on their own forks as we saw.

alext-mkrs:
I see. Well, we shouldn't be failing the build in case this is an external fork pull request and there's no variable set, because that's a legitimate case where it can fail - and we even have an "else" clause with `exit 0` for that, which never executes now.

At the same time we should fail it if variables are not set in a situation when it's a master branch push or an internal pull request. How about just moving the variable definition checks into a function and running the check for the first `if` and `elif` clauses of the PR situation check? That will allow us to remove this "allowed failure" thing, which now is too broad IMO and lets both genuine as well as expected failures slip through.

.travis.yml
alext-mkrs: I see. Well, we shouldn't be failing the build in case this is an external fork pull request and there're no variables set, because that's a legitimate case where it can fail - and we even have an "else" clause with `exit 0` for that, which never executes now.

At the same time we should fail it if variables are not set in a situation when it's a master branch push or an internal pull request. How about just moving the variable definition checks into a function and running the check for the `if` and `elif` clauses of the PR situation check? That will allow us to remove this "allowed failure" thing, which now is too broad IMO and lets both genuine as well as expected failures slip through. At the same time it doesn't require repo name hardcoding.

dnoliver: Agree with you, actually Mraa sonar scans where failing because there where no "sonar.jsava.binearies" required property, the build was marked as success. What I see are this requirements:

1. Remove Allowed Failure
2. intel-iot-devkit/master branch must always run sonar scan, and fail the build is it not pass. If sonar found to many bugs, this will not fail the build. That failure will be reported in the sonar dashboard
3. "Internal" PRs must always run sonar scan, and fail the build is it not pass. If sonar found to many bugs, the PR will be marked as failure. 
4. For external PRs, The sonar-scan.sh already skip the scan. We cannot run sonar scan for this because the required env vars are not shared with external PRs. We could create a temporary internal PR to make sonar run the scan and get a preview of the results (this will be manual)
5. For forks, check if required variables are set, and skip the scan with a message explaining that if there do not exists. Otherwise, follow the same policy for intel-iot-devkit/master branch and internal PRs

alext-mkrs: Yep, nicely summarized. And I think it's rather easy to achieve by just moving the variable checks inside the `if` branches (turning the check into a function to avoid code duplication). And to avoid hardcoding the "SLUG" name check, we can simply skip the scan, but not fail the build if variables are not set, assuming that if someone needs a scan, they'd define variables and those who need a scan, but forgot to set variables, would be attentive enough to check logs and see that the scan is actually skipped.

PR #798, opened 2017-08-10
Fix compiler warnings for ARM compilation

alext-mkrs:
I don't have actual ARM-based board to do a full testing, but the changes are obvious enough I think. If anyone in possession of a BBB or Phyboard or something else ARM is willing to test these, that'd be nice.

### DIFF
--- a/include/arm/am335x.h
+++ b/include/arm/am335x.h
@@ -63,9 +63,9 @@ mraa_result_t mraa_am335x_mmap_write(mraa_gpio_context dev, int value);
 /**
  * Unsetup register of mmap_gpio[]
  *
- * @return static mraa_result_t indicating success of actions.
+ * @return mraa_result_t indicating success of actions.
  */
-static mraa_result_t mraa_am335x_mmap_unsetup();
+mraa_result_t mraa_am335x_mmap_unsetup();
 
 /**
  * Read from gpio_context 'dev'
--- a/src/arm/96boards.c
+++ b/src/arm/96boards.c
@@ -39,6 +39,7 @@
 #define PLATFORM_NAME_DB410C "DB410C"
 #define PLATFORM_NAME_HIKEY "HIKEY"
 #define PLATFORM_NAME_BBGUM "BBGUM"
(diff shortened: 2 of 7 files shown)


alext-mkrs:
BTW, we should probably add compilation for various pre-set architectures into CI, to make sure we build those regularly, otherwise we always go over a default one and others may bitrot or at least get "dirty" like in this example of various warnings being there for a while.

dnoliver:
@alext-mkrs I can help with doing ARM builds in Travis CI. Do you have any guidance on how to do it?

Propanu:
@dnoliver, you'll have to setup a docker image with a cross-compiler. Check with @pylbert, as he already has a working solution for this, at least for UPM, that was never merged upstream.

Mani-Sadhasivam:
@alext-mkrs Tested the PR on 96Boards platform Dragonboard410c(ARM). A single warning gets generated for beaglebone.

`/home/linaro/mraa/src/arm/beaglebone.c:379:18: warning: variable ‘uart3_enabled’ set but not used [-Wunused-but-set-variable]
     unsigned int uart3_enabled = 0;`

You might want to clean that as well. Other than this the build looks clean :-)

alext-mkrs:
@dnoliver, what @propanu says would be the best. A simple, less than ideal, but still viable solution would be to add a line to existing configs, similar to the MOCK one. That will get this code path compiling each time at least (so we can capture such warnings or plain broken stuff).

@mani-sadhasivam, thanks a lot, that's very nice of you! That single warning is something I purposedly left out, as while we don't use that uart3 in Beaglebone, we may in the future, so I left that init code intact, even though it results in an unused variable. I've commented on this in the commit message BTW.

alext-mkrs:
@arfoll, @propanu, @whbruce, if there are no objections by Saturday, I'll merge this as I think we have enough confidence.

Propanu:
These changes look OK to me!

alext-mkrs:
Thanks @Propanu, merged.

PR #807, opened 2017-08-22
examples: Remove heap allocation from C++ examples

pylbert:
Small cleanup of MRAA C++ examples.  Switched from heap allocation to
stack allocation when possible.  This simplifies the samples since it
removes the need for explicit memory management.


pylbert:
Merged.

PR #811, opened 2017-08-29
Add on board LED support

Mani-Sadhasivam:
This patch adds support for using on board LED through sysfs.
Commonly available LED parameters are supported.


Mani-Sadhasivam:
@arfoll I've updated the PR will following changes:

1. Added close() api
2. Added C++ binding

I have left LED table as of now. Will add it once this PR gets in

Propanu:
@arfoll this looks good to me now and it would be a great addition for the next tagged version of mraa.

arfoll:
Merged, thanks!

PR #812, opened 2017-08-30
iio: Allow mraa_iio_trigger_buffer to handle void* args

pylbert:
Previously, mraa_iio_trigger_buffer took a void* args but did not use
this.  Included implementation to allow user to pass a void* to this
method and have the corresponding pointer returned in the interrupt
handler provided.

### DIFF
--- a/api/mraa/iio.h
+++ b/api/mraa/iio.h
@@ -102,7 +102,7 @@ mraa_iio_context mraa_iio_init(int device);
  * @param args Arguments
  * @return Result of operation
  */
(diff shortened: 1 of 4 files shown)


pylbert:
This change allows cleanup of UPM sensor examples.  The current UPM examples use global pointers to sensor objects.  This change enables handler routines to have a pointer to the original UPM sensor object so global pointers can be removed.

Unfortunately I don't have any of these sensors to test the added functionality on.  Maybe others can look over my changes or test with a kxcjk1013, mmc35240, l3gd20, or apds9930?

pylbert:
I verified the handler will be called using the mock platform and a few hacks to fake an iio device.

PR #816, opened 2017-09-07
Adding PWM support to Pi

ncrastanaren:
So, the Pi doesn't have userspace level PWM support, but you can mmap the raw registers and configure it that way, which is what this patch is doing. Tested with Raspberry Pi 3 B. I am able to set the duty cycle, change the frequency, and set the pulse width. Signals verified with logic analyzer.

Let me know if there's anything that you think needs cleaning up.

Note that the only accessible pwm is on GPIO18, which is also used by the audio subsystem.


arfoll:
Looks good to me! Merged with some commit message changes (just copied and pasted your github PR msg into the commit to get some more details).

PR #818, opened 2017-09-08
Up2 merge

nicolalunghiemutex:
This set of patches add support for UP Squared board from AAEON ( http://www.up-board.org/upsquared) and contain also a couple of fixes for the UP Board and mraa.


arfoll:
Changed the commit titles around a little but apart from that all merged! Thanks!

PR #819, opened 2017-09-11
docs: require specific version of doc tools

dnoliver:
Signed-off-by: Nicolas Oliver <dario.n.oliver@intel.com>


dnoliver:
According to what I understand from the Cmake find_package docs (https://cmake.org/cmake/help/v3.0/command/find_package.html), you must add an `EXACT` flag to force the exact version. In this case CMake will try to figure out compatible versions. 

But I actually thougth that CMake use semver to check lib compatiblity (like npm does), but apparently it use some cmake version files. 

Anyway, what do you think about removing the `PATCH` version number here? and just using i.e. `Sphinx 1.3`

arfoll:
You're right - cmake's version tag is not that nice actually :/ I'd suggest we simply hope for the best unless we know that only above a certain version will work like so:
```if (SWIG_VERSION VERSION_GREATER 3.0.4)```

I think this makes the most sense, we hope that versions above will work and we maybe have to have another IF statement if it doesn't. We already do thsi for cmake too. But we shouldn't try restrict newer versions just because semver changed, we should always try support the latest version of everything.

**Comment by dnoliver** [comment]
I made some tests changing yuidoc version and checking how it behaves

```
#### With Exact Version  works :)
root@e5dfc95b8c48:/usr/src/app# yuidoc --version
0.10.2

-- Found Yuidoc: /root/.nvm/versions/node/v6.11.0/bin/yuidoc (Required is at least version "0.10.2")
INFO - Yuidoc version is 0.10.2

#### Yuidoc Missing fails :)
root@e5dfc95b8c48:/usr/src/app# npm uninstall -g yuidocjs

CMake Error at /usr/share/cmake-3.5/Modules/FindPackageHandleStandardArgs.cmake:148 (message):
  Could NOT find Yuidoc (missing: YUIDOC_EXECUTABLE) (Required is at least
  version "0.10.2")

#### With lower patch version should not work :( but it works
root@e5dfc95b8c48:/usr/src/app# yuidoc --version
0.10.0

-- Found Yuidoc: /root/.nvm/versions/node/v6.11.0/bin/yuidoc (Required is at least version "0.10.2")
INFO - Yuidoc version is 0.10.0

#### With lower minor version must not work :( but it works
root@e5dfc95b8c48:/usr/src/app# yuidoc --version
0.10.0

-- Found Yuidoc: /root/.nvm/versions/node/v6.11.0/bin/yuidoc (Required is at least version "0.10.2")
INFO - Yuidoc version is 0.8.0
```

I was hoping to avoid the  IF statement to check for versions, and let Cmake handle it. But cmake is accepting older versions. I will check the docs a little more and add version checking manually if cannot make cmake handle this.

**Comment by dnoliver** [comment]
I had to do a manual check :( cmake does not handle versions properly if you are searching something with a custom find module .cmake file.

I made this change for yuidoc, and I am getting the expected behaviour. 
```
diff --git a/src/javascript/CMakeLists.txt b/src/javascript/CMakeLists.txt
index 118cf45..f0093a4 100644
--- a/src/javascript/CMakeLists.txt
+++ b/src/javascript/CMakeLists.txt
@@ -109,8 +109,8 @@ add_custom_command (TARGET npmpkg POST_BUILD COMMAND ${CMAKE_COMMAND} -E copy
 add_dependencies (npmpkg mraajs)

 if (BUILDDOC)
-  find_package (Yuidoc 0.10.2 REQUIRED)
-  if (YUIDOC_FOUND)
+  find_package (Yuidoc 0.10 REQUIRED)
+  if (YUIDOC_FOUND AND YUIDOC_VERSION VERSION_GREATER "0.10")
     add_custom_target (jsdoc ALL
       COMMAND ${CMAKE_SOURCE_DIR}/doxygen2jsdoc/docgen.js -m mraa -i xml/ -o jsdoc -c ${CMAKE_CURRENT_SOURCE_DIR}/doxygen2jsdoc_custom.json --strict
       COMMAND ${YUIDOC_EXECUTABLE} -C --no-sort --helpers ${CMAKE_SOURCE_DIR}/doxygen2jsdoc/generators/yuidoc/tmpl/mraa/helper.js --themedir ${CMAKE_SOURCE_DIR}/doxygen2jsdoc/generators/yuidoc/tmpl/mraa -o html/node jsdoc/yuidoc/mraa
@@ -120,7 +120,7 @@ if (BUILDDOC)
       COMMENT "Generating API documentation with Yuidoc" VERBATIM
   )
   else ()
-    message (STATUS "INFO - Failed to find Yuidoc. node.js API doc will not be generated")
+    message (SEND_ERROR "ERROR - Failed to find a compatible version of Yuidoc. node.js API doc will not be generated")
   endif ()
 endif ()
```

Cmake will fail and stop if YUIDOC_VERSION is below 0.10. Do you think this is the way to go?

**Comment by dnoliver** [comment]
This is how the log looks like now:

```
-- Found Doxygen: /usr/bin/doxygen (found suitable version "1.8.11", minimum required is "1.8") 
-- Found Sphinx: /usr/bin/sphinx-build (Required is at least version "1.3") 
-- Sphinx version is 1.3.6
-- Found Yuidoc: /root/.nvm/versions/node/v6.11.0/bin/yuidoc (Required is at least version "0.10") 
-- Yuidoc version is 0.10.2
```

arfoll:
@dnoliver Thanks alot, very thorough patch set - merged.

@Propanu the major node.js issues where that we care about the v8 API version and v8 versioning is a magic hex like but not hex string with a variable length because why not! Fun times....

PR #826, opened 2017-09-26
rpi: Add support for Raspberry Pi Zero W

skiselev:
This patch adds support and autodetection for Raspberry Pi Zero W.
It also adds more revisions for Raspbery Pi Zero from here:
http://elinux.org/RPi_HardwareHistory

Signed-off-by: Sergey Kiselev <sergey.kiselev@intel.com>


arfoll:
Looks good - merged. Would probably need to sort out the RPI platform and use platform_version.

PR #827, opened 2017-09-27
Enable force platform minnowboard turbot

tripzero:
WIP

Compile error would result if force minnowboard max. This is
because mraa_intel_minnowboard_byt_compatible signature requires
bool option for max or turbot board variations. This patch allows
user to force platform for either max or turbot without compile
error.


arfoll:
Uhm I can see why you do this but can we avoid adding something in types.h?
Also you need to sign off your patches.

tripzero:
We need some way of specifying the subtype for turbot.  The only other way I can think of is to create another CMAKE option for subtype.

I'm running into this because I'm trying to use mraa in a container.  I can get around it by forwarding /sys/devices/virtual/dmi/id/board_name into the container and letting autodetect work.

arfoll:
Why do we need a way of specifying the subtype? Since it's basically the same platform this seems odd...

PR #852, opened 2018-01-04
Topic/fix adc up1

nicola-lunghi:
Various fixes for aio
1) problem with max_analog_value:
if the adc has a resolution of 8 bit for example the maximum value is 255.
if the default resolution set for the aio module is 10 bit we get that a value of 255 get shifted by 2 bit and gives 1020 (not 1023)
the max analog value was before set to 2^10 -1 = 1023
so the maximum value of the adc would not give a floating value of 1 (1020 < 1023)
2) some values were set for every read but could be set in the init function
3) there's something that I don;t understand about the numbering of the pin in the aio modules.
the UP was getting a wrong ANALOG pin value of 28.
Fixed setting aio_non_seq in the up adc properties in the board definition variable

Regards
Nicola Lunghi

### DIFF
--- a/examples/analogin_a0.c
+++ b/examples/analogin_a0.c
@@ -60,6 +60,7 @@ main()
         adc_value_float = mraa_aio_read_float(adc_a0);
         fprintf(stdout, "ADC A0 read %X - %d\n", adc_value, adc_value);
         fprintf(stdout, "ADC A0 read float - %.5f (Ctrl+C to exit)\n", adc_value_float);
+        sleep(1);
     }
 
     r = mraa_aio_close(adc_a0);
--- a/src/aio/aio.c
+++ b/src/aio/aio.c
@@ -33,6 +33,7 @@
 #define DEFAULT_BITS 10
 
 static int raw_bits;
+static unsigned int shifter_value;
 
 static mraa_result_t
 aio_get_valid_fp(mraa_aio_context dev)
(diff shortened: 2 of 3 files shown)


nicola-lunghi:
Another question:
I still don't get the need to normalize adc values to a default of 10 bits.
Could not be better to have a function to query the ADC bit resolution and leave the raw value of the converter with his resolution??

https://github.com/intel-iot-devkit/mraa/blob/master/src/aio/aio.c#L144

nicola-lunghi:
signed off

arfoll:
Looks good - will merge asap

The reason it's all normalized to 10bit was to keep everything compatible with arduino's implementations in the galileo days. Whilst probably not so important, there may be some UPM modules that rely on that for some dodgy maths (@Propanu ideas?). My understanding was that 12bit adcs wheren't that precise anyways so 10bit was probably good enough... If someone feels strongly about just using whatever is there, we could change this in mraa 2.0 as long as it doesn't break too much of UPM.

arfoll:
Merged. thanks!

nicola-lunghi:
Thanks

Propanu:
I'm perfectly fine with this. There are a few analog sensors in UPM that have a hardcoded 1023 divisor, but I don't mind changing those to either use the `read_float()` or `1 << getBits() - 1` and keep it generic.

Issue #817, opened 2017-09-08
mraa_gpio_lookup return value


NickBridge:
Hi,

according to "mraa_gpio_lookup" function declaration in "common.h" file, return statement should be:

return i;

and not:

return plat->pins[i].gpio.pinmap;

Is it right?

alext-mkrs:
Hmm, looks indeed like a bug to me. @vintummala, could you please comment, as the author of that piece of code?

vintummala:
The intent of mraa_gpio_lookup() is to pass the name of GPIO pin and get the corresponding MRAA index number for it.

In peripheralman.c, we get the gpio names and assign the MRAA index number to pinmap for the names. This is done in mraa_peripheralman_plat_init() api.
Returning the same in mraa_gpio_lookup().

alext-mkrs:
I see, thanks. Well, it looks like this is in contradiction with what `pinmap` field is used everywhere else. This field normally contains the OS-level pin number, while something that we would call "mraa pin index" is namely the index within the pin table, i.e. `i` in this case. As an example - `mraa_gpio_init()` uses namely that as an input and `pinmap` is used only in the `_internal()` part of the init, to actually export the GPIO at the OS level.

I see that in `mraa_peripheralman_plat_init()` the value of `i` is being assigned to `pinmap` actually, but that is a sort of a corner case and doesn't really justify using `pinmap` as the "mraa pin index`.

In addition, just above the `mraa_gpio_lookup()` we have a reverse function - `mraa_get_pin_name()`, where again the input parameter is used namely as a pin table index, there's no search within `pinmap`. And that's a correct behavior.

So looks like we either need to create a proper `_lookup()` function, which would use the pin table index + make the current `_lookup()` Peripheral Manager-specific, or change the Peripheral Manager-related piece to actually use the pin index + correct the `_lookup()` to use it as well. @arfoll, any ideas?

alext-mkrs:
@arfoll, please? :)

alext-mkrs:
@arfoll, @propanu, what do you think of this one, given @vintummala's explanation and my investigation above?

Propanu:
Since we're talking about common functionality here, I would prefer to keep the `_lookup()` functions generic and not tied to a specific platform. In fact, even @vintummala agrees to this and I quote:

> The intent of mraa_gpio_lookup() is to pass the name of GPIO pin and get the corresponding MRAA index number for it.

[...]

PR #858, opened 2018-01-28
mraa.c: make mraa_gpio_lookup() return proper pin index

alext-mkrs:
Here goes the fix for #817, per our discussion there. I've checked and `mraa_gpio_lookup()` is actually not used anywhere in the mraa or upm code (including mraa's `peripheralman.c`).

This, in addition to the fact that Peripheral Manager-specific init code assigns mraa indexes to `pinmap` field, means to me that this change is going to be transparent to it and no changes needed there, as the values are effectively the same.

I'd leave it to those more familiar with ways Peripheral Manager-specific code uses this function, to decide if that's really the case, but to me it looks as if there's even no need to postpone merging this until the 2.0.

### DIFF
--- a/src/mraa.c
+++ b/src/mraa.c
@@ -883,7 +883,7 @@ mraa_gpio_lookup(const char* pin_name)
     for (i = 0; i < plat->gpio_count; i++) {
          if (plat->pins[i].name != NULL &&
              strncmp(pin_name, plat->pins[i].name, strlen(plat->pins[i].name) + 1) == 0) {
-             return plat->pins[i].gpio.pinmap;
+             return i;
          }
     }
     return -1;


vintummala:
This is just to clarify Alex comments:

Comment I' am trying to clarify is:
------------------------------------
 " **I've checked and mraa_gpio_lookup() is actually not used anywhere in the mraa or upm code (including mraa's peripheralman.c).**"

Explanation:
-------------
Peripheralman.c is the file through which MRAA can interact with Android Things Peripheral Manager. We can even write Android applications which interact with the peripherals through UPM and through MRAA and AT. mraa_gpio_lookup() is written to expose the pin numbers from the applications.
UPM drivers expect inputs to the constructor to be a MRAA number which application developers will not be aware of. The expectation here is in Applications we use the lookup function to get the pin number and then pass it to UPM. Please refer https://github.com/intel-iot-devkit/android-things-samples/
The applications in there uses the lookup function.

The change made in gpio_lookup looks good but I think this may need a change in peripheralman init() function.

alext-mkrs:
Thanks @vintummala. I've actually analyzed the init function before submitting the PR and it doesn't use the lookup function. Yes, it initializes pinmap fields with the mraa indexes, but namely due to that I believe that changing the lookup function return value doesn't affect anything.

I'll check those examples you mentioned to see if there's any additional context worth taking into account.

@propanu, thanks for approving the request. I'll double-check the examples and merge it if there's nothing affected over there.

alext-mkrs:
So I've analyzed those examples and I still think this change won't break anything, namely due to the way GPIO pins are initialized in `peripheralman.c`.

Here's why I think so:
1. In `peripheralman.c` we initialize GPIOs like this:
```c
//Updating GPIO bus structure
    for (; i < gpios_count; i++) {
        b->pins[i].name = gpios[i];
        //Retrieve this information from PIO once the API is available
        b->pins[i].capabilities = (mraa_pincapabilities_t){ 1, 1, 1, 1, 1, 1, 1, 1 };
        b->pins[i].gpio.pinmap = i;
}
```
i.e. `pinmap` field is set to be equal to the index in the pin array.
1. In the `mraa_gpio_lookup()` function, we returned the value of `pinmap` field before my change. After that change, we return the index (`i`) and they are numerically equal due to initialization described above.
1. AT examples use this function [like this](https://github.com/intel-iot-devkit/android-things-samples/blob/master/grovebutton/src/main/java/com/example/upm/androidthings/driversamples/GroveButton.java):
```java
gpioIndex = mraa.getGpioLookup(getString(R.string.Button_Edison_Arduino));
<...>
button = new upm_grove.GroveButton(gpioIndex);
```
where
```xml
<string name="Button_Edison_Arduino">IO0</string>
```
i.e.
* before the change that would be `mraa.getGpioLookup("IO0")` ==> return the `pinmap` value for the pin with the matching name;
* after the change that would be `mraa.getGpioLookup("IO0")` ==> return the `i` value for the pin with the matching name. `i` starts counting from 0, same as the `i` used to initialize the `pinmap` field value - so they are numerically equal;

Let me now pause for a couple of days to see if anyone else has any comments or objections and then I'll merge this.

alext-mkrs:
Merged with 2d210621ccc8377201e7301856d5102d254cd468

PR #861, opened 2018-01-30
Cleanup Python Examples

Mani-Sadhasivam:
This PR cleans-up the Python examples and fixes the Python docs.


Mani-Sadhasivam:
@arfoll Any update on this PR?

examples/python/gpio_advanced.py
```diff
@@ -21,34 +21,39 @@
...
 c = Counter()
 
 # inside a python interrupt you cannot use 'basic' types so you'll need to use
 # objects
-def test(gpio):
-  print("pin " + repr(gpio.getPin(True)) + " = " + repr(gpio.read()))
-  c.count+=1
+def isr_routine(gpio):
+    print("pin " + repr(gpio.getPin(True)) + " = " + repr(gpio.read()))
+    c.count+=1
```
alext-mkrs: While we are at it, this needs spaces around `+=`

examples/python/gpio_advanced.py
```diff
@@ -21,34 +21,39 @@
 # LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
 # OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION
 # WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE
+#
+# Example Usage: Reads GPIO 6 through interrupt
```
alext-mkrs: This description is IMHO a bit misleading. The idea of this functionality and the example is not in the specific fact we *read* the GPIO when the interrupt happens, but in the general fact is that we can *execute a predefined piece of code upon GPIO state change*, so I'd suggest highlighting namely that.

examples/python/i2c_bmp85.py
```diff
@@ -21,20 +21,21 @@
...
+# initialise I2C
 x = m.I2c(0)
 x.address(0x77)
 
 # initialise device
 if x.readReg(0xd0) != 0x55:
-  print("error")
+    print("error")
 
-# we want to read temperature so write 0x2e into control reg
```
alext-mkrs: I think this comment actually makes sense, so tha tpeople understand what the example does with the sensor, without looking into the datasheet. Why do you think it needs to be removed?

examples/python/led.py
```diff
@@ -0,0 +1,53 @@
...
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+# EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+# MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+# NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE
+# LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
+# OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION
+# WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+#
+# Example Usage: Reads maximum brightness value for user1 led and turns it
+# 	             on/off depending on current state. Then sets led trigger
```
alext-mkrs: The indentation is off on this line

examples/python/led.py
```diff
@@ -0,0 +1,53 @@
...
+
+# initialise user1 led
+led_1 = mraa.Led("user1")
+
+# read maximum brightness
+val = led_1.readMaxBrightness()
+
+print("maximum brightness value for user1 led is: %d" % val);
+
+if (val >= 1):
```
alext-mkrs: It would probably be helpful to clarify the logic of this piece in a comment.

examples/python/spi_loopback.py
```diff
@@ -21,20 +21,23 @@
...
+    txbuf = bytearray(4)
+    for y in range(0,4):
+        txbuf[y] = rand.randrange(0, 256)
 
+    # send and receive data through SPI
+    rxbuf = dev.write(txbuf)
+    if rxbuf != txbuf:
+        print("Data mismatch!")
+        break
+        exit(1)
```
alext-mkrs: Unreachable code. I know it was in the original example, but let's just fix that while we are at it. I'd suggest removing the `break` and leaving just the `exit(1)`, that would be in line with my fix for other example.

examples/python/uart_sender.py
```diff
@@ -21,6 +21,10 @@
 # LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
 # OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION
 # WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE
+#
+# Example Usage: Sends data through UART and expects `X` flag as response from
```
alext-mkrs: This description is incorrect, `_sender` doesn't expect to receive any flags, it *sends* a bunch of stuff, including a flag and then just reads and displays whatever data it receives after that.

examples/python/gpio_advanced.py
Mani-Sadhasivam: Okay. How about `Trigger ISR upon GPIO state change`?

examples/python/gpio_advanced.py
Mani-Sadhasivam: Okay. Will add spaces around `=` in all examples

examples/python/i2c_bmp85.py
Mani-Sadhasivam: Hmm. I thought these examples are here to _just_ demonstrate the basic usage and not provide much of the device specific details. However, I agree with you. Having this line will help the user without looking into datasheet. I will keep it as it is!

examples/python/led.py
Mani-Sadhasivam: Ack

examples/python/spi_loopback.py
Mani-Sadhasivam: Ack

examples/python/led.py
Mani-Sadhasivam: Okay. Will add a comment.

examples/python/uart_sender.py
Mani-Sadhasivam: Err. This was my bad. How about, `Sends some info messages along with X flag and expects a response from the other end. This example requires uart_receiver.py to be running on the other end`?

Mani-Sadhasivam:
@alext-mkrs Incorporated your review suggestions. Below is the changeset:

1. Squashed both commits into one
2. Added spaces around `=` and `+=`
3. Modified and added some comments based on review

Mani-Sadhasivam:
I'm not sure of the conflict appearing now :(

examples/python/gpio_advanced.py
alext-mkrs: Yep, sounds good.

alext-mkrs:
Those conflicts are because of my recent fixes in those files, you just need to rebase to the current master.

alext-mkrs:
@Mani-Sadhasivam, BTW, that conflict is not a problem, I can fix that by myself upon merging, but let me see if you'd want to train your git-fu by fixing that yourself ;)

Mani-Sadhasivam:
@alext-mkrs Rebased on top of master!

examples/python/spi.py
```diff
@@ -22,21 +22,24 @@
...
+txbuf = bytearray(3)
+txbuf[0] = 0x01
+txbuf[1] = 0x80
+txbuf[2] = 0x00
 
+while True:
+    # send data through SPI
+    rxbuf = dev.write(txbuf)
+    value = ((rxbuf[1] & 0x03) << 8) | rxbuf[2]
+    print value
```
alext-mkrs: Ok, this was one of the two conflicts between my fixes and your first version of the PR. You've picked your version here, but this reverts my fix - and will produces Sonar hits. Should be using `print()` form, like all other places we use this function.

Mani-Sadhasivam:
@alext-mkrs Done! Lack of sleep ;-)

arfoll:
Sorry was at FOSDEM2018 - this is great, merged.

PR #892, opened 2018-04-10
96boards: chardev: Move chardev enablement inside platform check

Mani-Sadhasivam:
Move the chardev enablement inside the platform check. Some boards might not have the chardev mapping yet.

### DIFF
--- a/src/arm/96boards.c
+++ b/src/arm/96boards.c
@@ -234,18 +234,21 @@ mraa_96boards()
             b->uart_dev[0].device_path = (char*) db410c_serialdev[0];
             b->uart_dev[1].device_path = (char*) db410c_serialdev[1];
             b->adv_func->gpio_mmap_setup = &mraa_db410c_mmap_setup;
+            b->chardev_capable = 1;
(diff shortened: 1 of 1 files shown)


Propanu:
Does [this](https://github.com/intel-iot-devkit/mraa/blob/master/src/mraa.c#L102) not work as intended?

Mani-Sadhasivam:
It works for platforms with latest kernel but not the ones without mapping. I should have updated the description. Anyway it's still better to enable for individual boards.

Propanu:
So I defined 3 cases for this:
* Platform doesn't have chardev mapping and doesn't list `chardev_capable = 1`, uses sysfs classes.
* Platform defines `chardev_capable = 1`, but kernel is too old. In this case the `gpiochip` search returns <= 0 and platform falls back to sysfs classes.
* Platform lists `chardev_capable = 1` and has devices, new code is used.

You should hit the 2nd one here. In any case, I do agree it's better to have individual boards list the capability separately.

malikabhi05:
Merged, Thanks!

PR #894, opened 2018-04-13
cmake: force c99 to be set in the compiler

arfoll:
Signed-off-by: Brendan Le Foll <brendan.le.foll@intel.com>


malikabhi05:
Merged, Thanks!

PR #896, opened 2018-04-17
gpio: chardev: fix leaking of gpiochip file descriptors

Mani-Sadhasivam:
The following APIs are only used for getting the line information from
kernel by opening the gpiochip independently and they fails to close
the file descriptor when done:

1. mraa_get_line_info_by_chip_number()
2. mraa_get_line_info_by_chip_name()
3. mraa_get_line_info_by_chip_label()

This will create issue if these API consumers like mraa_gpio_read_dir(),
mraa_gpio_mode(), mraa_gpio_chardev_dir() gets called in a loop. The
system will run out of file descriptor after some time.

Fix this issue by closing the opened file descriptors before freeing
the cinfo struct.


malikabhi05:
Merged, Thanks!

PR #913, opened 2018-07-26
ioinit: Update to build/test C++ ioinit

pylbert:
Multiple changes necessary to enable this.  The main goal is to build
and run the C++ ioinit unit tests (which require c++11)

    * Updated mraa required CMake version to 2.8.11 (this is needed for
      the target_xxx_xxx CMake syntax.
    * Added function for adding the c++11 flag give a CMake target.
    * Updated unit tests for range of CMake versions (tested on 2.8.11,
      3.8.2, 3.9.6, and 3.12.0).
    * Added C++ unit test file (more needed here).)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required (VERSION 2.8)
+cmake_minimum_required (VERSION 2.8.11)
 project (mraa C CXX)
 
 FIND_PACKAGE (Threads REQUIRED)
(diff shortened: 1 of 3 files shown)


pylbert:
Hey guys, can you take a look?  There are many ways to add the c++11 flag here.  I took an easy route.  We may want to add it to the pkgconfig file or create a second for building against the C++ headers.

I upped the CMake requirement, 2.8.11 provides target_include_directories.  In the long run, we may want to refactor mraa CMakeLists.txt files to remove global includes in favor of the target version.

pylbert:
I left the commented out tests as placeholders for @mihaitghstefanescu to fill in for c++ coverage.  I can remove them.

PR #914, opened 2018-07-31
CMake: Bump min version to 3.1

pylbert:
Removed function for setting c++11 standard in favor of using CMake 3.1
global:

set (CMAKE_CXX_STANDARD 11)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,5 +1,8 @@
-cmake_minimum_required (VERSION 2.8.11)
+cmake_minimum_required (VERSION 3.1)
 project (mraa C CXX)
+
+# Use the CMake 3.1 means to set the CXX standard
+set (CMAKE_CXX_STANDARD 11)
 
 FIND_PACKAGE (Threads REQUIRED)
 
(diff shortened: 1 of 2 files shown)


pylbert:
It seems like this should also be captured in the mraa.pc file however I'm not sure of the BKM for adding CXXFLAGS to pkg-config.  Variable maybe?

alext-mkrs:
I guess `CMAKE_CXX_FLAGS` would be the way - I haven't used it that much, but some quick googling hasn't revealed anything better than that.

arfoll (approved):
I still think dual .pc files is probably the only way here. But the best thing to do would be maybe to find out what other libs are doing before jumping to a solution :)

In the meantime Cmake 3.1 min is a good choice imho

pylbert:
I'm not going to get to this for a bit (moving). I will wrap it up asap
when settled.

alext-mkrs:
@pylbert, let me bump this up, do you still plan to merge this one?

alext-mkrs:
Closing this one without merging, as we've bumped minimum CMake version to 3.22 recently.

PR #923, opened 2018-08-30
Add support for initializing onboard LEDs based on board definition

Mani-Sadhasivam:
This patchset adds support for initializing onboard LEDs based on board definition. LED board definition has been added for 96Boards Dragonboard410c with 6 LEDs as an example.

For initializing LEDs based on the function name, a new API `mraa_led_init_raw` has been introduced. To accompany this change, `c` example is modified to pass LED index instead of name. C++, Python examples need not be modified since they have been mapped to `mraa_led_init_raw` API.

### DIFF
--- a/api/mraa/led.h
+++ b/api/mraa/led.h
@@ -47,15 +47,23 @@ extern "C" {
 typedef struct _led* mraa_led_context;
 
 /**
+ * Initialise led_context, based on led index.
+ *
+ *  @param led ID of the LED
+ *  @returns LED context or NULL
+ */
+mraa_led_context mraa_led_init(int led);
+
+/**
  * Initialise led_context, based on led function name.
  * The structure of LED entry in sysfs is "devicename:colour:function"
  * This api expects only one unique LED identifier which would be
- * "function" name most often. For instance, `mraa_led_init("user4");`
+ * "function" name most often. For instance, `mraa_led_init_raw("user4");`
  *
- *  @param led Name of the LED
+ *  @param led_dev Name of the LED device
  *  @returns LED context or NULL
  */
-mraa_led_context mraa_led_init(const char* led);
+mraa_led_context mraa_led_init_raw(const char* led_dev);
 
 /**
  * Set LED brightness
--- a/include/arm/96boards.h
+++ b/include/arm/96boards.h
@@ -38,6 +38,7 @@ extern "C" {
 #define MRAA_96BOARDS_LS_SPI_COUNT  1
 #define MRAA_96BOARDS_LS_UART_COUNT 2
 #define MRAA_96BOARDS_LS_PIN_COUNT  40
+#define MRAA_96BOARDS_LED_COUNT 6
 
 mraa_board_t* mraa_96boards();
 
--- a/include/mraa_internal_types.h
+++ b/include/mraa_internal_types.h
@@ -46,7 +46,7 @@
 #define MAX_AIO_COUNT 7
 #define MAX_UART_COUNT 6
 #define MAX_PWM_COUNT 6
-
+#define MAX_LED_COUNT 12
 
 // general status failures for internal functions
 #define MRAA_PLATFORM_NO_INIT -3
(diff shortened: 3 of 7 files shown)


include/arm/96boards.h
```diff
@@ -38,6 +38,7 @@ extern "C" {
 #define MRAA_96BOARDS_LS_SPI_COUNT  1
 #define MRAA_96BOARDS_LS_UART_COUNT 2
 #define MRAA_96BOARDS_LS_PIN_COUNT  40
+#define MRAA_96BOARDS_LED_COUNT	6
```
alext-mkrs: Some whitespace error between define name and value - I guess that's a TAB symbol instead of space

include/mraa_internal_types.h
```diff
@@ -454,6 +454,16 @@ typedef struct {
     /*@}*/
 } mraa_aio_dev_t;
 
+/**
+ * A Structure representing a LED device.
```
alext-mkrs: style: no need in the first "A" and the second one should be "an" to be aligned with the style you've used elsewhere

src/led/led.c
```diff
@@ -127,7 +127,43 @@ mraa_led_init_internal(const char* led)
...
+    }
+
+    dev = mraa_led_init_internal((char*) plat->led_dev[index].name);
+    if (dev == NULL) {
+        return NULL;
+    }
+
+    snprintf(directory, MAX_SIZE, "%s/%s", SYSFS_CLASS_LED, dev->led_name);
+    if (stat(directory, &dir) == 0 && S_ISDIR(dir.st_mode)) {
+        syslog(LOG_NOTICE, "led: init: current user don't have access rights for using LED %s", dev->led_name);
```
alext-mkrs: typo: "don't" -> "doesn't"

include/mraa_internal_types.h
```diff
@@ -490,6 +500,8 @@ typedef struct _board_t {
     mraa_adv_func_t* adv_func;    /**< Pointer to advanced function disptach table */
     struct _board_t* sub_platform;     /**< Pointer to sub platform */
     mraa_boolean_t chardev_capable;  /**< Decide what interface is being used: old sysfs or new char device*/
+    mraa_led_dev_t led_dev[MAX_LED_COUNT]; /**< Array of LED devices */
+    int led_dev_count; /**< Total onboard LED device count */
```
alext-mkrs: I'd suggest to use `unsigned int` here as we're unlikely to have a negative number of those. Otherwise please correct the check at line 141 in `led.c` to account for possible negative values and not only 0.

api/mraa/led.h
```diff
@@ -46,16 +46,24 @@ extern "C" {
  */
 typedef struct _led* mraa_led_context;
 
+/**
+ * Initialise led_context, based on led index.
+ *
+ *  @param led ID of the LED
+ *  @returns LED context or NULL
+ */
+mraa_led_context mraa_led_init(int led);
```
alext-mkrs: `unsigned int` would be safer choice for the parameter type, alternatively can be guarded by the runtime checks if you have any objections for unsigned.

include/arm/96boards.h
Mani-Sadhasivam: Ack

include/mraa_internal_types.h
Mani-Sadhasivam: Ack

src/led/led.c
Mani-Sadhasivam: Ack

include/mraa_internal_types.h
Mani-Sadhasivam: `unsigned int` makes sense here. Will change it

api/mraa/led.h
Mani-Sadhasivam: I'd be tempted to go for `int` and runtime check since passing the negative value would result in huge numbers and IMO it's good to report the proper error here.

alext-mkrs: Sure. I see you've added checks - ok with me.

PR #944, opened 2018-12-27
Update arm.c

bozkurthan:
To detect DE10-Nano with default config.

### DIFF
--- a/src/arm/arm.c
+++ b/src/arm/arm.c
@@ -73,7 +73,7 @@ mraa_arm_platform()
                     else if (mraa_file_exist("/sys/class/leds/green:ph24:led1")) {
                         platform_type = MRAA_BANANA;
                     }
-                } else if (strstr(line, "DE0/DE10-Nano-SoC")) {
+                } else if (strstr(line, "DE0/DE10-Nano-SoC") || strstr(line, "Altera SOCFPGA")) {
                         platform_type = MRAA_DE_NANO_SOC;
                 }
             }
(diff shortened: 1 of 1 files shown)


malikabhi05:
@bozkurthan  please add a sign-off on your commit.

src/arm/arm.c
```diff
@@ -73,7 +73,7 @@ mraa_arm_platform()
                     else if (mraa_file_exist("/sys/class/leds/green:ph24:led1")) {
                         platform_type = MRAA_BANANA;
                     }
-                } else if (strstr(line, "DE0/DE10-Nano-SoC")) {
+                } else if (strstr(line, "DE0/DE10-Nano-SoC") || strstr(line, "Altera SOCFPGA")) {
```
alext-mkrs: Please add this as a separate `else if` statement to align with the code style used here.

bozkurthan: I tried to fix problems that you mentioned. If something is wrong, let me know. Thanks for helping and being courteous. I accidentally added 1 wrong commit but I fixed.

alext-mkrs:
@bozkurthan thanks for the PR update. Looks like we're almost there - there are two things I'd like you to fix before I can merge. Let me know if you want me to do that instead, that's not a big deal, but I guess that might be a nice exercise for you if that's what you're interested in:
1. Please squash all three commits into one (and please don't forget to have the signoff in the resulting one). You can do that by using e.g. `rebase -i` and then force-pushing to the PR branch, the PR will update automatically;
1. Please correct the indentation for the code comment you've added - it's now off by 4 to the right.

tingleby:
Merged with 08b2cefebec8238bc8c025b259ec7f69fd0587c1
I went ahead and squashed the commits and fixed up the indenting.

bozkurthan:
Sorry for late action thanks for fixing.

PR #949, opened 2019-01-28
Mock platform

stefan-andritoiu:


tests/unit/api/api_common_h_unit.cxx
```diff
@@ -91,15 +91,15 @@ TEST_F(api_common_h_unit, test_libmraa_common_methods)
         /* No versioning for the mock platform */
         ASSERT_STREQ(NULL, mraa_get_platform_version(0));
 
-        /* Currently 10 pins in the mock platform */
-        ASSERT_EQ(10, mraa_get_pin_count());
+        /* Currently 11 pins in the mock platform */
+        ASSERT_EQ(11, mraa_get_pin_count());
```
alext-mkrs: Please combine all those test changes into the same commit that introduces the PWM pin - so that the pin addition goes in as an atomic change and we have both functionality and tests for it right away.

PR #952, opened 2019-02-05
CMakeLists.txt: Provide libmraa rpm packaging support

ghost:
Yocto recipes manage all packages as RPMs. This patch enables rpm
packaging for libmraa that can be used in the Yocto flow.

Signed-off-by: Sai Hari Chandana Kalluri <chandana.kalluri@xilinx.com>

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -214,6 +214,30 @@ if (BUILDDOC)
   endif (DOXYGEN_FOUND AND DOXYGEN_VERSION VERSION_GREATER "1.8")
 endif ()
 
+if (RPM)
+  message (STATUS "INFO - Enabled RPM packaging for ${DETECTED_ARCH}")
+  set(CMAKE_INSTALL_PREFIX "/usr")
+  set(CPACK_PACKAGE_VERSION ${VERSION})
+  set(CPACK_GENERATOR "RPM")
+  set(CPACK_PACKAGE_NAME "libmraa${mraa_VERSION_MAJOR}")
+  set(CPACK_PACKAGE_RELEASE 1)
+  set(CPACK_PACKAGE_VERSION
+      "${mraa_VERSION_MAJOR}.${mraa_VERSION_MINOR}.${mraa_VERSION_PATCH}${mraa_PACKAGE_ON_TAG}${VERSION_COMMIT}")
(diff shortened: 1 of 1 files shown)


tingleby:
Thanks for the PR! I am not a big user of Yocto but would you not normally package using 'make install' then the Yocto/Bitbake stage handle the RPM creation?

ckalluri:
@tingleby  The do_install task in bitbake copies the built files along with their hierarchy to locations that would mirror their locations on the target device. The installation process copies files from the work and build directories to the ${D} directory to create the structure as it should appear on the target system.

For the software to be installed correctly, it depends on how the software is built and the build system being used. If the Yocto recipe builds using Autotools or CMake, the Yocto OpenEmbedded build system understands how to install the software. Consequently, you do not have to have a do_install task as part of your recipe. However, if you wish to install additional files not already being installed by make install, you should do this using a do_install_append function using the install command. 

In short, Yocto thinks that libmraa which uses CMake already creates the RPMs. If the lines of code from the PR are not included, the RPMs will not be created in the Yocto flow.

arfoll:
Cpack isn't used in the cmake bbclass? Why would any of these change anything?

alext-mkrs:
> In short, Yocto thinks that libmraa which uses CMake already creates the RPMs. If the lines of code from the PR are not included, the RPMs will not be created in the Yocto flow.

@ckalluri, that's a bit strange, at least with the opkg packages (and to the best of my recollection DEB/RPM as well, though here's a bit more fuzzy) nothing additional was ever necessary. My understanding that it's namely as @tingleby described, Yocto creates RPMs on its own, not relying on the application's CMake targets beyond `install`, by doing the `install` and then collecting and packaging the produced files and dirs. Which Yocto version do you have and what mraa recipe(s) do you use?

tingleby:
Yeah it would seem what Id assume from experience with Debian packaging, using the Debian tool-set that the project itself doesn't need to be aware of Distro packaging.

I did a quick build of Poky and using the mraa recipe from `meta-oe/recipes-extended/mraa` I was able to get several RPMS from it :/

alext-mkrs:
Okay, thanks Tom. @ckalluri, please clarify why you need this, otherwise I'd personally be inclined to reject this for no good justification.

alext-mkrs:
Closing this one without merging, as per the past discussion, this should not really be necessary.

PR #953, opened 2019-02-16
BeagleBone: Fix detection of emmc and PCM for mraa pins 74-77

Chuckduey:
This is a fix for the beaglebone.c code that makes it so it can be run without the SD card on the BBG & BBB and use pins 74-77 as GPIO/SPI.  Several people have asked me for this code.  I thought it might be nice to share with the rest of the  group.

This fixes the problem of detecting if pins 74-77 are available for gpio/spi use.  These pins were sown to be used by the emmc, but actually used by the sound pcm.  If the sound PCM is disabled and the emmc is enabled the pinse can now be used.  mraa-gpio list output Pins 74-77
73     GPIO115: GPIO
74     GPIO113: GPIO SPI
75     GPIO111: GPIO SPI
76     GPIO112: GPIO SPI
77     GPIO110: GPIO SPI

to build:
1. clone repo
2. sudo apt-get install build-essential python-dev cmake automake libpcre3 libpcre3-dev byacc flex swig3.0
3. cd mraa, mkdir build, cd build
4. cmake -D CMAKE_INSTALL_PREFIX=/usr ..
5. make
6. sudo make install

This has been verified to work on BBGW, and BBBW with 9.0-9.5 with emmc enabled and PCM aduio turned off.  It is so nice to have control of these pins when running with out an SD card.

Signed-off-by: ChuckDuey <cduey@msn.com>

### DIFF
--- a/src/arm/beaglebone.c
+++ b/src/arm/beaglebone.c
@@ -370,6 +370,7 @@ mraa_beaglebone()
 {
     unsigned int emmc_enabled = 1;
     unsigned int hdmi_enabled = 1;
+    unsigned int pcm0_enabled = 1;
     unsigned int i2c0_enabled = 1;
     unsigned int i2c1_enabled = 1;
     unsigned int spi0_enabled = 0;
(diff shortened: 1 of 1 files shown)


tingleby:
Looks good to me. I don't have any recent experience with these boards
@michael-ring does this seem ok?

Chuckduey:
Thanks for the approval, there are not as many BB users out there, but I am sure they will find this very useful as some of my fellow BBers have.

tingleby:
Merged thanks!

PR #955, opened 2019-02-27
Typo in RTS signal for UART0

focalplane:


alext-mkrs:
@focalplane thanks for your correction, but yes, specifically, please add a sign-off line into your commit message and add an affected file/module name into the commit message summary. The link @Mani-Sadhasivam shared is the full information on that.

PR #965, opened 2019-05-02
gpio: Introduce mraa_gpio_init_by_name API

Mani-Sadhasivam:
This commit introduces mraa_gpio_init_by_name API for initializing
a GPIO by its line name provided by the kernel. This feature depends
on the GPIO chardev support and also the line names present in devicetree
or board files. Accessing GPIO using its line name, removes the dependency
from MRAA specific pin mapping and provides a cleaner way to access GPIOs.
This will solve the issue created by an external gpiochip probing before
the SoC's internal gpio controller and thereby making the MRAA pin mapping
wrong.

Currently, this API only supports initializing a single GPIO at a time. Bindings for the
API will come in successive PRs.

### DIFF
--- a/src/gpio/gpio.c
+++ b/src/gpio/gpio.c
@@ -164,6 +164,122 @@ init_internal_cleanup:
 }
 
 mraa_gpio_context
+mraa_gpio_init_by_name(char* name)
+{
+    mraa_board_t* board = plat;
+    mraa_gpio_context dev;
+    mraa_gpiod_group_t gpio_group;
+    mraa_gpiod_line_info* linfo = NULL;
+    mraa_gpiod_chip_info* cinfo;
+    mraa_gpiod_chip_info** cinfos;
+    int i, line_found, line_offset;
+
+    if (name == NULL) {
+        syslog(LOG_ERR, "[GPIOD_INTERFACE]: Gpio name not valid");
+        return NULL;
+    }
+
+    if (!board->chardev_capable) {
+        syslog(LOG_ERR, "[GPIOD_INTERFACE]: gpio_init_by_name not available for this platform!");
+        return NULL;
+    }
+
+    dev = (mraa_gpio_context) calloc(1, sizeof(struct _gpio));
+    if (dev == NULL) {
+        syslog(LOG_CRIT, "[GPIOD_INTERFACE]: Failed to allocate memory for context");
+        return NULL;
+    }
+
+    dev->pin_to_gpio_table = malloc(sizeof(int));
+    if (dev->pin_to_gpio_table == NULL) {
+        syslog(LOG_CRIT, "[GPIOD_INTERFACE]: Failed to allocate memory for internal member");
+        mraa_gpio_close(dev);
+        return NULL;
+    }
+
+    dev->num_chips = mraa_get_chip_infos(&cinfos);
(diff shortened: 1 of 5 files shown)


src/gpio/gpio_chardev.c
```diff
@@ -376,56 +379,85 @@ mraa_get_line_values(int line_handle, unsigned int num_lines, unsigned char outp
...
-                num_chips++;
-            }
-        }
-        closedir(dev_dir);
-    } else {
-        syslog(LOG_ERR, "[GPIOD_INTERFACE]: opendir() error");
+    int num_chips;
+    struct dirent** dirs;
+
+    num_chips = scandir("/dev", &dirs, dir_filter, alphasort);
```
Propanu: Requires glibc > 2.15 but I don't feel like a version check is explicitly required.

src/gpio/gpio.c
```diff
@@ -163,6 +163,122 @@ mraa_gpio_init_internal(mraa_adv_func_t* func_table, int pin)
...
+    }
+
+    /* Iterate over all gpiochips in the platform to find the requested line */
+    for_each_gpio_chip(cinfo, cinfos, dev->num_chips)
+    {
+        for (i = 0; i < cinfo->chip_info.lines; i++) {
+            linfo = mraa_get_line_info_by_chip_name(cinfo->chip_info.name, i);
+            if (!strncmp(linfo->name, name, 32)) {
+                /* idx is coming from `for_each_gpio_chip` definition */
+                syslog(LOG_ERR, "[GPIOD_INTERFACE]: Chip: %d Line: %d", idx, i);
```
alext-mkrs: This doesn't look like an error to me, so `LOG_ERR` should probably be replaced.

src/gpio/gpio.c
```diff
@@ -163,6 +163,122 @@ mraa_gpio_init_internal(mraa_adv_func_t* func_table, int pin)
...
+
+    /* Save the provided array from the user to our internal structure. */
+    dev->provided_pins = malloc(dev->num_pins * sizeof(int));
+    if (dev->provided_pins == NULL) {
+        syslog(LOG_CRIT, "[GPIOD_INTERFACE]: Failed to allocate memory for internal member");
+        mraa_gpio_close(dev);
+        return NULL;
+    }
+
+    memcpy(dev->provided_pins, &line_offset, dev->num_pins * sizeof(int));
```
alext-mkrs: Could `num_pins` ever be bigger than `provided_pins` here? If yes there's a risk of buffer overflow.

src/gpio/gpio.c
Mani-Sadhasivam: ah yes, this is a debugging clutter.. will remove it

src/gpio/gpio.c
Mani-Sadhasivam: It will not. But how that matters? `provided_pins` allocation depends on `num_pins` so eventually, both sizes will be the same.

alext-mkrs: Didn't have time yesterday to fully trace the relation between those two, so decided just ask to speed things up :smiley: Now I see that `provided_pins` is indeed allocated based on `num_pins`, so no problem here.

alext-mkrs:
LGTM now, feel free to merge.

PR #966, opened 2019-05-02
[WIP] CMake Modules revisit

tingleby:
As part of the move to the Eclipse Foundation, we are reviewing the IP details included. 
This PR removes some old modules in favour of built in features of cmake


tingleby:
I've naturally tried it on x86 also on an arm64 vm. I need to dig out an arm board and try it out before pulling the trigger. 

`CMAKE_SYSTEM_PROCESSOR` is key. Yocto seem to handle it well, http://git.yoctoproject.org/cgit/cgit.cgi/poky/tree/meta/classes/cmake.bbclass#n86

@Mani-Sadhasivam what do you think?

Propanu:
This is awesome, I will pull it to UPM too since it will save the Eclipse folks quite a bit of headache.
@Mani-Sadhasivam why not use containers for that?

tingleby:
@Mani-Sadhasivam cross compiling should be done via CMAKE directives like a toolchain file.
https://cmake.org/cmake/help/v3.6/manual/cmake-toolchains.7.html#cross-compiling-for-linux

Mani-Sadhasivam:
> @Mani-Sadhasivam cross compiling should be done via CMAKE directives like a toolchain file.
> https://cmake.org/cmake/help/v3.6/manual/cmake-toolchains.7.html#cross-compiling-for-linux

@tingleby Yeah but user shouldn't need to modify cmake environment of MRAA, we should provide a cmake flag like BUILDARCH for cross compiling.

tingleby:
But it needs to know what compiler/linker/sysroot etc to use. One flag would not be enough.

You can even in your environment setup a cmake wrapper for cross compiling. 
mingw-w64-cmake from the Arch User Repo provides something cool like that, 
https://aur.archlinux.org/cgit/aur.git/tree/mingw-cmake.sh?h=mingw-w64-cmake#n15

Mani-Sadhasivam:
> But it needs to know what compiler/linker/sysroot etc to use. One flag would not be enough.
> 

Yes I agree but we should be able to point to the compilers/toolchain path supported by most of the distros. If the user wants something special(then he shouldn't be a novice) he can always go and customize it based on the need. Just a thought!

My point was to provide an easy to use interface for users without modifying much code. If it doesn't make enough justification, we should at least document the instructions.

PR #1024, opened 2020-06-13
CMakeLists.txt: add ENABLECXX option

ffontaine:
C++ is a mandatory dependency since version 1.4.0 and https://github.com/eclipse/mraa/commit/122cab1f1e53b7c9c7cd82905b962071f9bad9dc

As a result, build on embedded toolchains that do not support C++ fails
on:

```
CMake Error at CMakeLists.txt:2 (project):
  The CMAKE_CXX_COMPILER:

    /home/naourr/work/instance-1/output-1/per-package/mraa/host/bin/arm-linux-g++

  is not a full path to an existing compiler tool.
```
Fixes:
 - http://autobuild.buildroot.org/results/31086422e03611c16ab59c4418e3669b580bc0c0

Signed-off-by: Fabrice Fontaine <fontaine.fabrice@gmail.com>

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,5 +1,5 @@
 cmake_minimum_required (VERSION 2.8.11)
-project (mraa C CXX)
+project (mraa C)
 
 FIND_PACKAGE (Threads REQUIRED)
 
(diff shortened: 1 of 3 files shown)


Propanu (commented):
I don't see any harm in adding this as an option, a few nitpicks though:
* the related UPM project has a similar option but it's called `BUILDCPP`, would it be possible to use the same name here?
* the CXX flags for warnings were missed, there may be a few more in the language config files as CXX is used during SWIG builds too, can we include those?

Thanks!

Propanu (approved):
@ffontaine thanks for making the changes, these can be merged as part of the next project release.

Issue #937, opened 2018-11-06
Error for gpio read on UP squared


yoann-solana:
Hi,
I am trying to read value on a pin with interruption but the read function is returning a negative value and not 0 or 1 as expected.

# Setup
* The board is an Intel Up squared running Ubuntu 16.04 with kernel 
```
$ uname -a  
Linux upsquared-UP-APL01 4.10.0-42-generic #5000~upboard11-Ubuntu SMP Mon Jun 25 15:58:51 UTC 2018 x86_64 x86_64 x86_64 GNU/Linux
```
I have followed all the [procedure](https://wiki.up-community.org/Ubuntu) to setup the UP.

To install MRAA, I followed the procedure from the [Readme](https://github.com/intel-iot-devkit/mraa#installing-on-ubuntu). The version is 2.0 (returned by mraa_get_version()).

# The code
The code (at the bottom) read the value of a pin on an interruption.During the reading I got an error. I chose the pin MRAA no.13 according to the [mapping](https://github.com/intel-iot-devkit/mraa/blob/master/docs/up2.md). I also tried with gpio 15, 16, 18 but still with errors. Also I am executing the code as a super user with sudo -E -S since it is a ROS node.

# Error
In the Interrupt Service Routine, I am using the mraa_gpio_read() function but it returns a negative value (error) and not 0 or 1. 
The log show an error when the mraa is initialized: 
*imraa: Failed to open lock file*
```
journalctl -t libmraa
-- Logs begin at mar. 2018-11-06 13:10:12 CET, end at mar. 2018-11-06 13:29:07 CET. --
nov. 06 13:29:07 th09001-PC libmraa[2538]: libmraa version v2.0.0 initialised by user 'root' with EUID 0
nov. 06 13:29:07 th09001-PC libmraa[2538]: Adding i2c bus found on i2c-5 on adapter i2c_designware.1
nov. 06 13:29:07 th09001-PC libmraa[2538]: Adding i2c bus found on i2c-4 on adapter i2c_designware.0
nov. 06 13:29:07 th09001-PC libmraa[2538]: up2: kernel pinctrl driver available
nov. 06 13:29:07 th09001-PC libmraa[2538]: imraa: Failed to open lock file
nov. 06 13:29:07 th09001-PC libmraa[2538]: gpio: support for chardev interface is activated
nov. 06 13:29:07 th09001-PC libmraa[2538]: libmraa initialised for platform 'UP2' of type 16
```
[...]

PR #1030, opened 2020-07-22
Allow GPIO ISR and state read on UP2

AndreasAZiegler:
This PR addresses #937.

By setting ```chardev_capable = 0``` it is possible to activate ISR on a pin AND read the state of the pin out.

### DIFF
--- a/src/x86/up2.c
+++ b/src/x86/up2.c
@@ -101,7 +101,7 @@ mraa_up2_board()
     b->platform_version = PLATFORM_VERSION;
     b->phy_pin_count = MRAA_UP2_PINCOUNT;
     b->gpio_count = MRAA_UP2_GPIOCOUNT;
-    b->chardev_capable = 1;
+    b->chardev_capable = 0;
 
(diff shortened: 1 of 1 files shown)


Propanu:
Hi @AndreasAZiegler, unfortunately this will also reduce GPIO read/write performance so my take is somehow it has to be offered as an option instead. I'll add a comment with more details on #937, I think what you're really asking for is for AAEON to update their pinmuxing driver.

AndreasAZiegler:
@Propanu Thanks for your information.
Would you mind, pointing me towards the place(s) where it is best to add such an option. Furthermore, do you know if it makes sense to report this to AAEON respectively is this already reported?

Propanu:
There were definitely some attempts to enable interrupts with the pinctrl driver Emutex provided by the community, but it's very likely you'll need to build it from source to get kernel ISR support for the board. Here are some links for the driver and the discussion I mentioned:
* https://github.com/emutex/ubilinux-kernel/tree/upboard-4.9/drivers
* https://lwn.net/Articles/769017/
* https://forum.up-community.org/discussion/3188/solved-kernel-driver-for-gpio

PR #1071, opened 2021-07-13
aio: Update shifter_value and max_analog_value on mraa_aio_set_bit

jan-kiszka:
mraa_aio_set_bit() changes value_bit, so the calculations of
shifter_value and max_analog_value become outdated. Move their
initialization to mraa_aio_set_bit and call that function from
mraa_aio_init instead.

Based on original patch by Le Jin.
@johnson325


tingleby (approved):

tingleby:
Thanks @jan-kiszka

PR #1078, opened 2021-10-23
Contrib i2c-0_1 (when i2c-1 exists without i2c-0)

g-vidal:
Building mraa on raspberry Pi 4 with bullseye  (kernel 5.*) leads to non-functional i2c bus because `i2c-0` does not exist any more on standard images.  Default i2c has become `i2c-1`; we propose here to extend the initial condition for creating i2c resources in `mraa.c` to the existence of **i2c-0 OR i2c-1** which may not interfere with other builds.
This trick can help until a cleaner solution is found.


tingleby:
Looks fine to me. Do you have any reference we could document to the default moving to `i2c-1` or is it just this platform.

g-vidal:
I have not checked why the i2c-0 was removed, it is beyond my competencies or require a lot of time I do not have at the moment. It might be a change in the default options of compilation that get the default i2C-0 pins for another purpose. I contributed this because if it happens with the build I am using it may occur for others...I'll come back if I find some information.

PR #1102, opened 2022-12-05
gpio: Fix JS binding regarding interrupt injections

jan-kiszka:
According to https://github.com/libuv/libuv/discussions/3847 and based on stress tests, it is not correct to call uv_queue_work outside of the loop thread. We rather need to use the async API of libuv. That even simplifies things.

Resolves "uv__queue_done: Assertion `uv__has_active_reqs(req->loop)' failed" errors that were easy to trigger by multiple DIs being used in parallel. See also https://github.com/siemens/meta-iot2050/issues/386.

Set to draft status as more testing might still be needed, including at our users. But early feedback would definitely be welcome!

### DIFF
--- a/api/mraa/gpio.hpp
+++ b/api/mraa/gpio.hpp
@@ -31,10 +31,10 @@
 #include <stdexcept>
 
 #if defined(SWIGJAVASCRIPT)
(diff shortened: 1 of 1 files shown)


jan-kiszka:
Looks good from out tests now.

PR #1103, opened 2023-01-12
aio: Fix setting adc bits when initializing

AsuraZeng:
Set the adc bits to the default value of 10,
so the raw bits of the platform will not take
effect during initialization, and it is always the default value

Signed-off-by: chao zeng <chao.zeng@siemens.com>

### DIFF
--- a/src/aio/aio.c
+++ b/src/aio/aio.c
@@ -13,8 +13,6 @@
 #include "aio.h"
 #include "mraa_internal.h"
 
-#define DEFAULT_BITS 10
-
 static int raw_bits;
 static unsigned int shifter_value;
 static float max_analog_value;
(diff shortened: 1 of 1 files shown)


jan-kiszka (changes requested):
The fix appear correct to me, but you should rework the commit message: set the number of bits to those that the hardware reports - rather than hard-coded 10 bits.

Also, DEFAULT_BITS seems unused now.

AsuraZeng:
> The fix appear correct to me, but you should rework the commit message: set the number of bits to those that the hardware reports - rather than hard-coded 10 bits.
> 
> Also, DEFAULT_BITS seems unused now.

yes, remove the `DEFAULT_BITS` and rewrote the commit.

jan-kiszka:
On second look, the situation looks less clear to me again: `raw_bits` is actually a global variable and not a local var that was forgotten to be used in `mraa_aio_set_bit` as I initially thought. So, it should be perfectly fine to first set a smaller value_bits by calling `mraa_aio_set_bit` and then let node-red-nodes raise that again to the value the concrete platform supports. The actual bug must be elsewhere, this here is just papering over it.

jan-kiszka:
None-issue, see https://github.com/siemens/meta-iot2050/issues/407. MRAA chose to start AIO pins with a hardware-independent number bit of bits, even if the hardware can do more. Nothing to do here.

PR #1108, opened 2023-06-06
rockpi4: fixed the issuse 'No Pins'

nascs:


RadxaYuntian:
@tingleby can you take a look at this PR? We are the device manufacturer, and we have a few more products that we'd like to add to MRAA. We intend to provide pre-built MRAA library in our systems. Thanks.

tingleby (approved):

PR #1112, opened 2023-08-22
rock3: Add initial support for radxa's rock3 family

jaysonsantos:
Hi there, I'd like to get some coaching if possible, to upstream the changes to radxa's rock3 family.
My steps were, I got a copy of rock4pi's code and from the gpio docs on the official website + dts from from a running armbian kernel I changed the code to adapt.


nascs:
Hi @jaysonsantos, this pr is only for Radxa ROCK 3A, Radxa ROCK 3A does not have the same gpio pinout as Radxa ROCK 3B and Radxa ROCK 3C.

RadxaYuntian:
FYI an implementation of ROCK 3C is available at #1113.

PR #1124, opened 2023-11-27
Add support for the new IOT2050-SM variant and some improvements

huaqianli:
Followed up on https://github.com/eclipse/mraa/pull/1102.

 1. iot2050: Add support for the new IOT2050-SM variant
    
    IOT2050-SM board was added into IOT2050 series, in contrast to existing
    variants, the IOT2050-SM board comes without a Arduino interfaces.

2. iot2050: add debugfs pinmux support
    
    This patch adds support for multiplexing pins via debugfs rather
    than access memory mapped pad-configuration registers.
    The debugfs pinmux offers the possbility to run mraa on iot2050
    platforms as regular user instead of root by adjusting privileges
    on debugfs files.
    
    Bias settings are currently also configured by accessing pinmux.
    Unfortunatelly a proper upstream-like pinconf usage is currently
    not possible.
    
    Note: In case debugfs mux fails MRAA falls back to mmap mux.

3. common: increase pin name size
    
    Some pin names are longer than 12 characters.
    32 characters should be enough while consuming not too
    much space.

### DIFF
--- a/api/mraa/common.h
+++ b/api/mraa/common.h
@@ -32,7 +32,7 @@
 /** Max size off Mraa Platform name */
 #define MRAA_PLATFORM_NAME_MAX_SIZE 64
 /** Size off Mraa pin name */
-#define MRAA_PIN_NAME_SIZE 12
+#define MRAA_PIN_NAME_SIZE 32
 
 /** Bit Shift for Mraa sub platform */
 #define MRAA_SUB_PLATFORM_BIT_SHIFT 9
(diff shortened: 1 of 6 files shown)


bniedermayr:
@huaqianli: Some fixes where pushed meanwhile [here](https://github.com/benbrenson/mraa/tree/devel/bennie/pinmux-simple) regarding debugfs due to device tree changes of the iot2050.
There are still issues related to non-root access though, since these changes are still relying on some sysfs functionalities.

... so there is still some work todo...

PR #1141, opened 2025-07-09
cmake: Set minimum required version to 3.5 for CMake 4+ compatibility

alperak:
Fix:

| CMake Error at CMakeLists.txt:1 (cmake_minimum_required):
|   Compatibility with CMake < 3.5 has been removed from CMake.
|
|   Update the VERSION argument <min> value.  Or, use the <min>...<max> syntax
|   to tell CMake that the project requires at least <min> but has been updated
|   to work with policies introduced by <max> or earlier.
|
|   Or, add -DCMAKE_POLICY_VERSION_MINIMUM=3.5 to try configuring anyway.
|
|
| -- Configuring incomplete, errors occurred!

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required (VERSION 2.8.11)
+cmake_minimum_required (VERSION 3.5)
 project (mraa C)
 
 FIND_PACKAGE (Threads REQUIRED)


alext-mkrs:
Thank you for your contribution. I think we can go further than that and there's some additional adjustment needed along with the version bump, so we'll tackle this in #1148. I can't close the PR, so you could do that (or @tingleby maybe you could do the honors).

tingleby:
https://github.com/eclipse/mraa/pull/1148 replaces this

PR #1148, opened 2025-11-02
cmake: bump minimum version to address a deprecation warning

alext-mkrs:
This fixes the following warning shown while configuring:
```
[cmake] CMake Deprecation Warning at CMakeLists.txt:1 (cmake_minimum_required):
[cmake] Compatibility with CMake < 3.10 will be removed from a future version of
[cmake] CMake.
```

We don't have a clearly defined supported distro/versions list, so I took Ubuntu 22.04 as arguably the oldest one we'd want to support for the reference point. It has CMake 3.22, released in July 2022. Other opinions on the choice of version are welcome.

There are additional warnings regarding policy 00148 (related to Python search logic), but I'll tackle those later - they may be a bit more involved and I want to setup build/test workflow first.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,15 +1,9 @@
-cmake_minimum_required (VERSION 2.8.11)
+cmake_minimum_required (VERSION 3.22)
 project (mraa C)
 
 FIND_PACKAGE (Threads REQUIRED)
 
-if (CMAKE_VERSION VERSION_LESS "3.1")
-  if (CMAKE_C_COMPILER_ID STREQUAL "GNU")
-    set (CMAKE_C_FLAGS "-std=gnu99 ${CMAKE_C_FLAGS}")
-  endif ()
-else ()
-  set (CMAKE_C_STANDARD 99)
-endif ()
+set (CMAKE_C_STANDARD 99)
 
 option (BUILDCPP "Enable C++ (needed by FTDI4222 and tests)" ON)
 


arfoll (approved):
I think the req for minimum 2.8.11 is from Poky 1.5 because we where not compatible with Poky 1.4 that shipped with galileo v1...

alext-mkrs:
Thanks Brendan (and long time!). There's indeed this legacy set of images for which I'm assuming that anyone still using those boards today and willing to build the newer mraa will have an updated CMake, in the interest of moving on.

OTMH, Edison is well-covered with the latest image available being on Yocto Scarthgap, Galileo is not so much, but I'll see what could be done if that becomes a problem (let us know by submitting an issue if you're reading this!).

arfoll:
> Thanks Brendan (and long time!). There's indeed this legacy set of images for which I'm assuming that anyone still using those boards today and willing to build the newer mraa will have an updated CMake, in the interest of moving on.

Hi! I certainly wasn't suggesting you should keep it! Was my way of saying the reason for keeping it so compatible probably doesn't exist any more (or shouldn't!).

> OTMH, Edison is well-covered with the latest image available being on Yocto Scarthgap, Galileo is not so much, but I'll see what could be done if that becomes a problem (let us know by submitting an issue if you're reading this!).

YP5.0 with an edison? Just found github.com/edison-fw... cool stuff.

arfoll (approved):

tingleby (approved):

